PR #134, opened 2022-05-18
Spelling fixes

supercrafter100:
This pull request contains a few spelling mistakes I have found in the new website.


shiffman:
Fantastic, thank you!!!

shiffman:
> I'm unsure where `NVideo RTX Broadcast` is from? I can only find `NVIDIA Broadcast App`? Referencing [this line](https://github.com/CodingTrain/thecodingtrain.com/blob/main/content/pages/faqs/how-livestream-and-record.json#L7)

Yes! NVIDIA Broadcast App is correct!

shiffman:
Seems like you are finding more and more! Let me know when your review is complete and then I will merge!

supercrafter100:
I've noticed that the file "mantain-website-guide" is misnamed and likely outdated since it looks like the things mentioned in it are not correct anymore. It will probably need to be partially rewritten to apply to the new website so i've stopped further spelling fixing in it for now.

supercrafter100:
I'm actually going to undo the renaming of the file since i'm unsure if it will break things (likely will). Someone else will have to fix that one.

supercrafter100:
Think i'm done here! I've read through most things now. Will create another PR if I find another issue later on maybe. Thanks for waiting!

shiffman:
❤️

PR #139, opened 2022-05-20
Fixed Reference box size.

Heinrich-XIAO:


src/components/LinkList.module.css
```diff
@@ -15,9 +15,10 @@
 .icon,
 .description,
 .url {
-  height: var(--baseline);
+  /* height: var(--baseline); */
   padding: 0 var(--spacing-normal);
   line-height: var(--baseline);
+  height: min-content;
```
fdoflorenzano, line 21: I see a couple problems here. Even though for links this works because the line-height is set to be a the baseline, a couple lines later in this style sheet the descriptions get a smaller one, making them not adjusted to the overall baseline of the site.

So I would go this way: Make the links behave the same way as descriptions currently do. Use just a third of the baseline as line height, which gives them more space to overflow and fits better to their font size.

Also, but not that important, you should remove all commented line.

fdoflorenzano (changes requested):
Left an idea of how to fix it!

shiffman:
Hi @Heinrich-XIAO any updates or thoughts on this? Can you share some screenshots of the issue if it is still a problem on the website? I'm not sure how to reproduce it. Thank you!

Issue #132, opened 2022-05-13
Layout with many many topics


shiffman:
Just noting some layout wonkiness now that I've added so many topics (tags) in! Maybe we should bring in the "show more" feature to the track pages too?

https://codingtrain-preview.netlify.app/tracks/code-programming-with-p5-js

<img width="1591" alt="Screen Shot 2022-05-13 at 10 29 08 AM" src="https://user-images.githubusercontent.com/191758/168305687-dbf4c933-c1ff-4aac-a791-94a04f75d666.png">

fdoflorenzano:
That could be a good idea! I can try the show more approach first and see how it feels.

Issue #129, opened 2022-05-12
Thumbnail Aspect Ratio on About Page


shiffman:
Just noting that we should probably update the thumbnail aspect ratio on the about page to match the challenges page.

About page:
<img width="1670" alt="Screen Shot 2022-05-12 at 3 05 18 PM" src="https://user-images.githubusercontent.com/191758/168149852-31f38540-a57e-4402-89cd-183291728255.png">

<img width="1640" alt="Screen Shot 2022-05-12 at 3 06 11 PM" src="https://user-images.githubusercontent.com/191758/168149940-f5b9cd26-4450-4349-8c2f-146083ecc60a.png">

Also, @fdoflorenzano should the titles on the challenges page be bolded like they are on the about page? I think that might look nice but defer to you all!

fdoflorenzano:
I agree the aspect ratio should match. I had a similar thought about it.

About the bolded titles, I'm not sure. There's too things happening there, the titles are bold and black, which brings a lot of contrast. This all also relates to color contrast concerns that I have, so I would probably focus on that aspect separately.

shiffman:
> About the bolded titles, I'm not sure. There's too things happening there, the titles are bold and black, which brings a lot of contrast. This all also relates to color contrast concerns that I have, so I would probably focus on that aspect separately.

Got it! I don't have a preference and trust you all on your design decisions! I just had noticed it was different so pointed it out, the aspect ratio is the more important note here!

PR #141, opened 2022-05-21
Visual tweaks and fixes

fdoflorenzano:
This PR fixes a couple issues noticed by @shiffman and me:
- #132 : Added a "show more" button for long tag lists, left the track header section overflow when expanded, and moved around some sections.
- #129 : Made the thumbnail layout consistent to challenges cards, and made a similar fix to guide cards.
- Buttons had CSS transitions setup to all CSS properties, which made them animate unintentionally when resizing the page. This fixes that behavior and leaves the transition to the intentional background animation.
- Fixed a little URL issue when clicking the titles in challenges cards in homepage. It was trying to go to tracks/ instead of challenge/

### DIFF
--- a/src/components/Button.module.css
+++ b/src/components/Button.module.css
@@ -70,7 +70,7 @@ button.rainbow.red {
   );
   background-size: 20000% 100%;
   background-position: right bottom;
-  transition: all 1s ease-out;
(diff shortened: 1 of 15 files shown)


fdoflorenzano:
@shiffman yes!

PR #145, opened 2022-06-01
Update passenger-showcase-guide.md

kfahn22:
When there are no contributions yet, there will not be a "showcase" folder.  I have edited the guide to clarify that the file name will need to include the showcase folder name for the first contribution.


shiffman:
Fantastic, thanks for this!

PR #156, opened 2022-06-04
Ulam spiral contributions

kfahn22:
I have ported all the existing contributions and added pngs (all except one that was just code) and added one that I did a while ago.  (The image is the one I am adding.)
contribution10


shiffman (commented):
@kfahn22 wow! Thank you! Noted one error causing the tests to fail. (Didn't check if there were others but I'll keep looking!)

kfahn22:
I can't figure out the error.  What do I need to do to fix (and avoid in future).

content/videos/challenges/167-prime-spiral/showcase/contribution10.json
```diff
@@ -0,0 +1,10 @@
+
+Raw Blame
```
shiffman, line 2:
```suggestion
```

kfahn22:
I thought it would be more efficient to do all at once, but maybe not if an error in one file will cause all to fail.

shiffman:
I was able to fix the error with https://github.com/CodingTrain/thecodingtrain.com/pull/156/commits/871fe30a4bd2ae7c4528478736c622fa226848c7. The contributions don't show in the preview so I cleared cache and ran the deployment again. Waiting to check!

shiffman:
It worked! Great job @kfahn22!!!

Screen Shot 2022-06-05 at 5 00 17 PM

Issue #44, opened 2022-01-18
Tests for content structure and file content


fdoflorenzano:
This is currently being worked on in the `content-struture-tests` branch.

We are considering thus far:

- [x] Testing the folder structure inside `content`
- [x] Testing file presence in each folder inside `content`
- [x] Testing JSON properties for all relevant files are present and follow expected structure
- [x] Testing that all mentioned slug strings correctly reference other files

shiffman:
Noting here that one of the tests is failing (I thought I filed an issue about this before, but I can't find it? Maybe we discussed on slack?)

<img width="1348" alt="Screen Shot 2022-06-04 at 3 23 58 PM" src="https://user-images.githubusercontent.com/191758/172022696-c392f344-448d-45a2-be04-30c92917b093.png">

shiffman:
Can this issue be closed? Tests are working great! We can adjust them or file new issues as necessary.

fdoflorenzano:
Good to close for me!

PR #157, opened 2022-06-05
fix: allows subfolders inside of guides and about content folders

fdoflorenzano:
Little fix to [comment](https://github.com/CodingTrain/thecodingtrain.com/issues/44#issuecomment-1146672822) that allows subfolders in `content/pages/guides`, and also `content/pages/about` because it seemed appropriate.


shiffman (approved):
Great!

PR #158, opened 2022-06-05
fix: sublists don't have left border

fdoflorenzano:
I noticed in the new guides a little style bug, lists get a left border, even when they are nested:
Screen Shot 2022-06-04 at 21 23 11

This PR fixes that:
Screen Shot 2022-06-04 at 21 23 32


shiffman:
Thank you!!

Issue #140, opened 2022-05-21
Minor UI adjustments 


raj-jaiswal:
While surfing through the website,I found these things that shouldn't take long to adjust.

On the landing page :

![Screenshot_20220521-064805_Chrome](https://user-images.githubusercontent.com/59106789/169629481-0b5491e1-1936-4356-8b95-c4a51c2ed9d3.png)

The YouTube membership option should start below the title "Support" unless intentional.

![Screenshot_20220521-065605_Chrome](https://user-images.githubusercontent.com/59106789/169629556-46502f86-1fd4-4db0-aeba-6a350fd07713.png)

The getting started,videoes and passengers area doesn't feel centered or aligned to the text above

On mobile version :

![Screenshot_20220521-065120_Chrome](https://user-images.githubusercontent.com/59106789/169629642-acacd492-00b5-4ebb-8f99-fb36e29761ae.png)

Here,The margin of the text is inconsistent (maybe it shouldn't wrap).

![Screenshot_20220521-065048_Chrome](https://user-images.githubusercontent.com/59106789/169629726-bf4c2d53-b9c3-4e27-b2c2-92b66c41757d.png)

In the support area,the white spacing seems a bit too much, making the buttons a bit more wider should help.

![Screenshot_20220521-065108_Chrome](https://user-images.githubusercontent.com/59106789/169629761-d92b19c4-75bc-4038-a140-e86e76642f30.png)

There seems to be no margin on images which feels a bit wrong. My suggestion would be to make the images at same margin as text or adding a global padding to the container.

On the coding challenges page :

![Screenshot_20220521-065313_Chrome](https://user-images.githubusercontent.com/59106789/169629849-ed04fe4c-5d70-4d9c-abfa-a6415584a801.png)

The contrast here is a bit less which makes it hard to read. Also,python is repeated 2 times.

On the tracks section :

![Screenshot_20220521-070104_Chrome](https://user-images.githubusercontent.com/59106789/169629947-91d2fae5-b83a-4b1b-a7f4-6874ee4ef05c.png)

This one might be noticed already but the description is bleeding.

fdoflorenzano:
> On the landing page :
> The YouTube membership option should start below the title "Support" unless intentional.

That's intentional.

> 
> The getting started,videoes and passengers area doesn't feel centered or aligned to the text above

If I understood what you mean right, that's also intentional.

> On mobile version :
[...]

Issue #121, opened 2022-05-09
Review front-end design and lighthouse report


fdoflorenzano:
## From Dana's Feedback
- [x] All aboard container. I would suggest to add the horizontal light lines.
- [x] All aboard container. I would suggest to align the text “welcome to the coding train with…” in the way that we had it set in our website. (more to the left)
- [x] About page: I think the main text should be in the same typeface that we have for header text. The “now that you met gloria” can stay in the monospace.. maybe?
- [x] Text in general: Line height and text size seems to vary in different pages, I would suggest to follow the tracks page for line height, size and color.
- [x] Text in general: 15 px
- [x] Text in general: Dark grey (If it doesn’t create issues with accessibility, I think it look better with the rest of the colors)

## From Fernando's Review
- [x] #131
- [x] Guide card layout is off
- [x] Challenge card layout in homepage
- [x] Buttons animate on resize too
- [x] Issues in top bar menu in screen sizes.

## From lighthouse report
- [x] Color contrast
- [x] Performance in certain pages

## Dan
- [x] #129
- [x] #130
- [x] #132

PR #159, opened 2022-06-05
Site level text revision and other tweaks

fdoflorenzano:
This PR mainly reviews the text styling across the site, and also tweaks some other aspects along the way:
- All gray text is darkened, making it more readable and makes the site a bit less bright across. As proposed in #121 
- Footer logos also get darkened.
- The default font size is set to 15px, also as proposed in #121 
- Makes padding and line-height more consistent across the site. Paragraphs share padding and line-height, as so do card layouts. For card consistency it also implied altering other text attributes. This was also proposed in #121 
- Tweaks CTA element in Footer to behave like a block, to fix padding and overflow bug noted in #140.
- Tweaks menu element alignment and organization, as a test proposed in #163. 
- Fix bug for smaller screen sizes where nav bar would get over pages and cut them off.
- Fix bug that added extra padding in between elements of the track stops in video pages.


shiffman (approved):
This looks great! Wasn't sure if it's still in progress so go ahead and merge when you are ready @fdoflorenzano!!!

PR #162, opened 2022-06-05
adding spring forces contribution

kfahn22:
I created a simple template (in a repo called showcase)  to add the contributions locally and then upload the entire folder.  I used the Mandelbulb and Ulam spiral folders as my guide, so I hope the format is correct.  If not, please let me know what I need to fix.

The index.json file needs some additions:  I wasn't sure what the url should be for the processing code  example and the image file for the folder is missing.

There is one contribution I did not include because I got a 404.


shiffman:
Hi @kfahn22! I went through and made some adjustments on this page:

1. p5 and Processing versions no longer need to be separate code listings, they fall under one example just with a "p5" and a "processing" url. I need to set up a new Coding Challenge example code repo to host the Processing ports. For now, we can link to the old "website" repo if need be. I just took them out here though.
2. relatedChallenges won't work if the video has not been ported yet. So the pendulum challenge would have to be ported for it to be listed as related to this one.
3. Lots of the data coming from the old site uses paths to videos on the old site. Those should be replaced with links to the path on the new site (if it exists) or just the YouTube video directly.

shiffman:
Noticing a couple things this pull request needs:

1. Thumbnail image for the challenge itself. (can be the YT thumbnail)
2. replacement of placeholder text for "Description of. . ." in the reference and example tabs

shiffman:
Weirdly, GH won't let me add an image b/c this originated in your fork (or at least I can't figure out how to do it.) Here's the image if you want to add it as `index.jpg` alongside `index.json`. I can also merge this and add it after the fact.

https://img.youtube.com/vi/Rr-5HiXquhw/maxresdefault.jpg

kfahn22:
How do I add the image?  I can grab a screen shot, but I am not sure how to get the actual png file from YouTube.

I have signed up to add  the pendulum coding challenges so I will keep the above in mind when I do so.

shiffman:
The link in my previous message is the thumbnail image! But don't worry about it. Let's merge this and we can add separately another time!

Issue #168, opened 2022-06-06
Thumbnail Images for Code Examples don't seem to work


dipamsen:
From https://github.com/CodingTrain/thecodingtrain.com/issues/68#issuecomment-1043285167, there is supposed to be "thumbnail images" for code examples like follows:

```jsonc
"codeExamples": [
  {
    "title": "Monty Hall",
    "description": "Three-door demonstration of the Monty Hall Problem with gameplay statistics",
    "image": "codeExample1.png",
    "urls": {
      "p5": "https://editor.p5js.org/codingtrain/sketches/pLW3_PNDM"
    }
  },
  // ...
]
```

## Actual Behaviour

- No thumbnail image showing up
  
   ![image](https://user-images.githubusercontent.com/59444569/172196133-4998bf8f-b210-45a6-826b-ae56c7eff864.png)

- It takes `codeExample1.png` image as the main thumbnail, instead of using `index.jpg`

   ![image](https://user-images.githubusercontent.com/59444569/172196394-9fdcf6ce-384b-4c8e-bad2-53466595ac8a.png)

(`codeExample1.png` contains test image, `index.jpg` contains actual video thumbnail)

fdoflorenzano:
And how are the files located in the file structure?

dipamsen:
I just assumed that all files are supposed to be in the root folder. Please lmk if they are supposed to be in a separate folder.

```
content/
  challenges/
    170-monty-hall-problem/
      index.jpg
      codeExample1.jpg
      index.json
```

shiffman:
Ah, yes, investigating this a bit more it looks like the example images are supposed to go in a subfolder called `images`, you can see this here:

https://thecodingtrain.com/tracks/code-programming-with-p5-js/code/2-variables/4-random
https://github.com/CodingTrain/thecodingtrain.com/tree/main/content/videos/code/2-variables/4-random

Can you try that and see if it works? 

Also, we should add info about this to the site maintenance guide (if it's not already there!)

dipamsen:
Thanks for the clarification.

Unfortunately, I still can't make the images work. 


File Structure:
```
content/
  challenges/
    170-monty-hall-problem/
      images/
        codeExample1.png
      index.json
```

Code:
```jsonc
"codeExamples": [
    {
      "title": "Monty Hall",
      "description": "Three-door demonstration of the Monty Hall Problem with gameplay statistics",
      "icon": "codeExample1.png",
      "urls": {
[...]

PR #171, opened 2022-06-06
Challenge thumbs fix

fdoflorenzano:
This PR fixes little typo bug reported in #168. Images and examples were loading correctly, just the linking between them had a little typo for the challenges.  

I added a random image to the Monty Hall challenge so you can check that in fact it's fixed.


shiffman:
Yes, this works! Thank you @fdoflorenzano!!!

PR #175, opened 2022-06-07
Color contrast fix

fdoflorenzano:
This PR worked on top of the progress in #159, so it should be reviewed after it. 

This PR addresses come accessibility concerns we noticed regarding color contrast, specifically in sections where color text was on top of colored background, as mentioned in #121 and #140.

We've come up with this solution, which replaces colored text on colored background with the same dark gray added in #159. It works well with the rest of the site, it introduces enough contrast level and it doesn't introduced new shades of colors to work with.


shiffman (approved):
This looks great I'll let you merge!

PR #179, opened 2022-06-07
Self avoiding walk

kfahn22:
Adding images for example sketches.


kfahn22:
I have edited the index.json file.  (I misspelled recursion in the image file so I misspelled in the index.json file but I can fix that if it bothers you.

shiffman:
> I have edited the index.json file. (I misspelled recursion in the image file so I misspelled in the index.json file but I can fix that if it bothers you.

Wonderful! I might fix it someday if I come across while doing other updates but let it not stop us now from merging!!!

shiffman:
Not sure why this one is stuck on the checks, starting it over!

PR #180, opened 2022-06-07
158 shape classifier

kfahn22:
Adding 158-shape-classifier challenge.

I wasn't able to get an image for the contribtution or the processing sketch.


shiffman:
> Adding 158-shape-classifier challenge.
> 
> I wasn't able to get an image for the contribtution or the processing sketch.

No worries! Maybe it makes sense to open an issue referencing this pull request and noting that the thumbnails need to be made. This will help for tracking and anyone else who wants to jump in and add them!

PR #181, opened 2022-06-07
Add YT Thumbnails

dipamsen:
added some missing thumbnails


shiffman:
> Yes, I just think that the thumbnails for the "newer" challenges are better suited on the website as most of the time they display the algorithm/output as they main focus. On the other hand, for some of the older challenges it will be better to use the algorithm image. Just my opinion :)

Sounds great to me!!!

PR #198, opened 2022-06-08
Ascii

kfahn22:
I have added the images for the example code, updated the index.json file to reference the images, and added the contributions.


shiffman:
Wow! Heroic effort @kfahn22. Noting for the future (this happened to me on the livestream as well) that sometimes a zoomed in crop of a code example looks better than the full canvas since they appear so tiny on the site.

Screen Shot 2022-06-08 at 1 13 49 PM

PR #203, opened 2022-06-08
[WIP] Discord redirect

fdoflorenzano:


shiffman:
Oooh! Awesome, I checked and we have these other redirects from the old site.

https://github.com/CodingTrain/website-archive/tree/main/_redirects

In theory, I could also maybe create a big table of redirects from the old challenge and "track" urls to the new ones but I think I'll just let google eventually sort it all out?

I do need to do a big pass of updating the YouTube descriptions to point to new website URLs at some point soon!

fdoflorenzano:
@shiffman I just added the other redirects to pointed out to!

Issue #170, opened 2022-06-06
Missing Meta Images


M0nica:
It looks like the head of the website is missing a meta images to render in previews when the link is shared. https://github.com/CodingTrain/thecodingtrain.com/blob/main/src/components/Head.js If it's confirmed that this is an issue, are there pre-made images that should be used for the meta images?

<img width="595" alt="Screen Shot 2022-06-06 at 3 20 37 PM" src="https://user-images.githubusercontent.com/6998954/172231990-8d54d07f-5022-4644-b93d-d5f0049c8001.png">
Screenshot from https://metatags.io/

shiffman:
Oh, this is a great point! I'm also thinking I might like to make some minor edits to the meta sentence that's included.

All of the track and video pages have a thumbnail image associated with them (usually called `index.png` or `index.jpg`) would it make sense to use this when available? Otherwise we could probably design a single preview image it could be the train image on the homepage or something else. @jasontheillustrator might have some ideas! @fdoflorenzano let us know if you have thoughts too!

Should we also have a favicon?

fdoflorenzano:
Hey! The referenced component should be easily extendable so support this. Using that and the Layout component, if the page has an associated thumbnail, use that as meta image, and if not use a default image as meta.

shiffman:
This can be closed with #207 I believe! I'll wait a bit in case @M0nica wants to double-check the implementation and how it looks now!

M0nica:
Great, it looks like https://github.com/CodingTrain/thecodingtrain.com/pull/207 resolves this.

PR #207, opened 2022-06-09
Meta images

fdoflorenzano:
This completes the metadata settings initially set for the site, and also adds images to the setup as proposed in #170.

The approach is taking both text and image content already worked on for every site and using it as metadata for each page.

For example, for titles and descriptions, which most pages have, this are also injected as metadata for the corresponding page. So tracks titles and descriptions get injected in the first video page of each track, and each individual video page also has their own title and description metadata.

Images would work the same way. I'll show here how each preview looks, and indicate where is the image taken from. Some images probably will need to be replaced, since some are still placeholders from a while ago.

Also a little note: Because the meta tags only accept full URLs for image sources, metadata images point to images that are available in `thecodingtrain.com`, instead of the current preview URL. Because there's consistency across builds, it should link to the same images if a new build has the same images as the main build. But if new images are added in a separate branch-preview, it probably won't point to the right image unless the branch is merged into main.

Path: `/`
Image: `content/tracks/main-tracks/placeholder.jpg` (This is also the default value for all pages. If there's no image available, this will show).

Screen Shot 2022-06-09 at 15 38 11

Path: `/tracks`
Image: `content/tracks/main-tracks/placeholder.jpg`
Screen Shot 2022-06-09 at 15 46 47

Path: `/challenges`
Image: `content/videos/challenges/placeholder.jpg`

Screen Shot 2022-06-09 at 15 47 15

Path: `/guides`
Image: `content/pages/guides/placeholder.png`

Screen Shot 2022-06-09 at 15 48 29

Path: `/about`
Image: First image linked in the key `"covers"` in `content/pages/about/index.json`

Screen Shot 2022-06-09 at 15 49 24

Path: `/challenges/#-specific-challenge`
Image: `content/videos/challenges/#-specific-challenge/index.jpg`. If that one doesn't exists, it fallback to `content/videos/challenges/placeholder.jpg`.

Screen Shot 2022-06-09 at 15 52 07

Path: `/tracks/specific-track`
Image: `content/tracks/[main|side]-tracks/specific-track/index.jpg`. If that one doesn't exists, it fallback to the videos image.

Screen Shot 2022-06-09 at 15 55 04

Path: `/tracks/specific-track/specific-video`
Image: `content/videos/specific-path/index.jpg`. If that one doesn't exists, it fallback to `content/videos/placeholder.png`.

Screen Shot 2022-06-09 at 15 58 22

Path: `/guides/specific-guide`
Image: `content/pages/guides/placeholder.png`

Screen Shot 2022-06-09 at 15 59 59

### DIFF
--- a/gatsby-config.js
+++ b/gatsby-config.js
@@ -1,7 +1,9 @@
 module.exports = {
   siteMetadata: {
     siteUrl: 'https://www.thecodingtrain.com',
-    title: 'The Coding Train'
+    title: 'The Coding Train',
+    description:
+      'All aboard the Coding Train with Daniel Shiffman, a YouTube channel dedicated to beginner-friendly creative coding tutorials and challenges.'
   },
   flags: {
     DEV_SSR: true
--- a/src/components/Head.js
+++ b/src/components/Head.js
@@ -2,14 +2,16 @@ import React, { memo } from 'react';
 import { Helmet } from 'react-helmet';
 import { useStaticQuery, graphql } from 'gatsby';
 
-const URL = 'https://thecodingtrain.com/';
-const defaultTitle = 'The Coding Train';
-const defaultDescription =
-  'All aboard the Coding Train with Daniel Shiffman, a YouTube channel dedicated to beginner-friendly creative coding tutorials and challenges.';
-
 const Head = ({ title, description }) => {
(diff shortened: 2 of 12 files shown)


shiffman (approved):
This is fantastic, thank you @fdoflorenzano!

PR #233, opened 2022-06-10
Local storage

kfahn22:
Adding a sketch that shows how to use local storage to share data across multiple WEBGL canvases.
Since the GitHub pages hosting and local storage were both in this folder, I wasn't sure about the best way to add a contribution.  Hopefully, this is OK
contribution1
.


kfahn22:
I am guessing this failed because the file structure is different for the tips and tricks.

shiffman:
@kfahn22 showcase submissions have to be associated with a specific video, not a track. So I think you unfortunately just have to pick one!

kfahn22:
Unfortunately, there really isn't a good place to put it.

PR #235, opened 2022-06-10
Coding Challenge 36 - Blobby

skaser85:


shiffman (approved):
This is great thank you @skaser85!!! For some of the challenges that have a nice visual, we don't necessarily need to use the YT thumbnail (esp if its an older design). But I can always update these later too if I want to do a "default image" pass over all the challenges at soem point!

PR #236, opened 2022-06-10
Coding Challenge 52 - Random Walker

skaser85:


shiffman:
This is fantastic, thank you @skaser85! A few small notes:

1. I see you pulled the topics from YouTube tags. Those take a "throw all the spaghetti at the wall" approach I think we can edit them down and just use a few consistent key "topics" across the challenges.
2. You are always welcome to add links to references, other videos, and the "related challenges" properties if you like!
3. You are also welcome to add timecodes (but I know that's a lot of work so not required).

Mostly I just want to point out that each port is an opportunity to clean up and improve the metadata!

I am so appreciative of everyone's generosity here in helping with the site! Let me know if you prefer I merge or if you'd like to make further changes!

skaser85:
Thanks for the feedback, Dan.  I can definitely improve things, so I think we should hold off on merging this.

shiffman:
This looks great now @skaser85! Just didn't want to merge prematurely, let me know if it's all set!

skaser85:
Thanks, @shiffman.  Yes, I think it's ready to merge now.

PR #238, opened 2022-06-10
Lissajous

kfahn22:
Adding 116-lissajous coding challenge


shiffman:
Hi @kfahn22! If you use a tool like https://github.com/yt-dlp/yt-dlp or even a website like https://youtube-thumbnail-grabber.com/ you can get a higher quality version of the YT thumbnail! 

Also, for some of these, a nice screenshot of the code example can work as the default image as well! I'll merge just noting for future reference, thank you AGAIN!

PR #240, opened 2022-06-11
Acknowledgements

shiffman:
@fdoflorenzano would it be possible to have an additional property for each person, so instead of:

```json
        {
          "name": "Editing and Video Production: Mathieu Blanchette",
          "url": "https://mathieu.com"
        },
```

We'd have:

```json
        {
          "name": "Mathieu Blanchette",
          "url": "https://mathieu.com",
          "role": "Editing and Video Production"
        },
```

Then the layout could look more like the following (with the url only linking to the name?)

Screen Shot 2022-06-11 at 1 28 10 PM

### DIFF
--- a/content/pages/about/index.json
+++ b/content/pages/about/index.json
@@ -44,12 +44,6 @@
   "featuredTitle": "Featured",
   "featured": [
     {
-      "title": "Image to ASCII (Coding Challenge 167)",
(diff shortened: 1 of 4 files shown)


fdoflorenzano:
@shiffman just pushed here the addition of role!

PR #248, opened 2022-06-11
Fix date ordinal numbers for special case

### LINKED ISSUE #247: Incorrect Ordinal Number for Dates
image
This should say Jun 11th 2022.

The code defaults to `st` when the date ends in 1, `nd` for 2 and `rd` for 3. 11, 12 and 13 are special cases which have to be accounted for.

dipamsen:
Fixes #247


shiffman:
This looks good to me but thought I would check with @fdoflorenzano for a review!

fdoflorenzano (approved):
✅

PR #250, opened 2022-06-11
Spirograph

kfahn22:
Adding 61-fractal spirograph.


content/videos/challenges/61-fractal-spirograph/index.json
```diff
@@ -0,0 +1,52 @@
+{
+    "title": "Fractal Spirograph",
+    "description": "In this coding challenge, I attempt to create a \u201cFractal Spirograph\u201d with Processing (Java). The visuals are inspired by images and explanation on C. J. Chen\u2019s blog.",
+    "videoNumber": "61",
+    "videoId": "0dwJ-bkJwDI",
+    "date": "2017-02-28",
+    "languages": ["processing", "java"],
+    "topics": [
+        "fractal spirograph",
+        "creative coding",
```
shiffman, line 10: I don't think we need a creative coding topic, it's maybe too broad and applies to almost everything?

content/videos/challenges/61-fractal-spirograph/index.json
```diff
@@ -0,0 +1,52 @@
...
+    "timestamps": [],
+    "codeExamples": [
+        {
+            "title": "Fractal Spirograph",
+            "description": "The sketch creates a fractal spirograph.",
+            "image": "spirograph.png",
+            "urls": {
+                "p5": "https://editor.p5js.org/codingtrain/sketches/Q60oWddzr",
+                "processing": "https://github.com/CodingTrain/Coding-Challenges/tree/main/061_fractal_spirograph/Processing",
+                "other": "https://github.com/CodingTrain/Coding-Challenges/tree/main/061_fractal_spirograph"
```
shiffman, line 24: Same question re: other here? Maybe we've been doing this so the download link shows up and I just didn't notice?

kfahn22: I am not sure that it  is necessary, butI believe that this is the pattern from the example you referenced (recursion), and  also in the template so I have been including it.

PR #256, opened 2022-06-12
Cardioid

kfahn22:
Adding 133-cardioid


kfahn22:
I checked the deploy log and there  errors are related to Image.png files.  Do you know what is going on?
Screen Shot 2022-06-12 at 10 08 08 AM

klinegareth:
It looks like you haven't uploaded an image named `pointcloud.png`. Once you upload that it should work, unless I'm missing something!
image

shiffman:
Hmmmm, looks like something did go wrong somewhere! The same tests are failing with my recent pull request #258 too! I'm investigating now but leaving shortly to have dinner with @runemadsen maybe we can sort it out before our entree comes. 😂

kfahn22:
I have reduced the file size for four  of the contribution images.

PR #257, opened 2022-06-12
Passenger Showcase files Cleanup 

dipamsen:
Cleaning up showcase files
- Removed extra 'by' from author name (originated from scraped template data, as `by <author-name>` was in a single html tag in the old website)
- Removed empty `author.url` fields


shiffman:
Great! Yes, I've been trying to find this option for me as the maintainer to merge with main but I can't seem to!

Issue #263, opened 2022-06-13
Spherical Geometry Challenge


shiffman:
I'd welcome a port of the spherical geometry challenge since I plan to refer to it in my PCD Copenhagen presentation tomorrow! Also, if anyone wants to add timecodes to the Super Ellipse challenge that would be a welcome addition!

shiffman:
Thank you @skaser85 and @kfahn22! Your are Coding Train heroes!

PR #265, opened 2022-06-13
added timestamps for the 19-superellipse coding challenge

skaser85:
see issue #263 - added timestamps


shiffman:
The build process should be back and running, can you merge this with main and it should re-run the tests and a new deploy preview! I'm not sure if it's an easy thing to do from your end so if it's not let me know and I'll figure it out!

PR #267, opened 2022-06-13
Update index.json

kfahn22:
I was checking links and released the link to the  polar coordinates video was broken.


shiffman:
Good catch thank you!

PR #268, opened 2022-06-14
Gatsby V4

runemadsen:
This PR upgrades the website to Gatsby V4. An overview of the changes:

- The most important thing is that the website now uses Node 16. **Please delete your local `node_modules` folder, upgrade to Node 16, and run `npm i` again. I added a `.nvmrc` file which states the exact version of Node, and Netlify will use this during the build process.
- Deleted unused Gatsby cloud plugin
- Deleted deprecated `DEV_SSR` flag

Everything seems to be building nicely on my computer, so let's see what happens on the Netlify PR build. We should check it real good before merging in!


shiffman:
Thanks! I'm running `v17.3.0` I assume I don't need to downgrade? I'm keeping my eye on the Netlify logs! 🤞

runemadsen:
@shiffman I think you can stay on `v17` :) 

Also, I ran a `du -a content | sort -n -r | head -n 20` in the repo to list the largest folders, and they entirely line up with the slow query warnings in the build. All of those folders have images that are larger than 1MB. I think that's the next thing to optimize: Figure out how to bring those images down a little bit.

runemadsen:
But wow, this is taking forever to build.

shiffman:
Yeah, this is the logs 😭

Screen Shot 2022-06-14 at 12 18 57 PM

runemadsen:
@shiffman So those flags were optional flags in Gatsby v3 but are enabled by default in v4. I think we can drastically cut down on the build time if we downsize some of the many showcase images. Many of them are 1, 2, 3 even 4 MB, and these are being resized on every build. Still poking around, so let's see if we can get it to at least build successfully.

shiffman:
Ah, right, I didn't consider the showcase images! There are zillions of them! Maybe I can do a quick PR to resize all of them down and then we'll need to implement some size specs in the guide, etc. (A reason to have a form too which can limit or reize!)

runemadsen:
@shiffman I think that's a good idea. I just pushed a commit where I downsized and PNG optimized the three largest showcase folders, so I'll see how that affects things. No matter what, it should take *that* long to resize, but if we have lots of large images, it make things much slower.

runemadsen:
There is also a feature on Netlify where it defers the resizing of the images to their CDN. We could look into this. I think it comes for free with any plan. I'll keep you posted on my progress.

runemadsen:
@shiffman I also updated the Netlify plugin on this current build, and now the logs say this (which I'm guessing is not bad):

Screen Shot 2022-06-14 at 12 41 53 PM

runemadsen:
Sorry for all the spam, but it is also a big suspect that I can build the site on my own computer in 5 minutes and the Netlify build takes more than 30m.

kfahn22:
Are you able to efficiently downsize the images on your end?

runemadsen:
Okay @shiffman: I managed to get the build to sorta succeed by downsizing and optimizing 20 of the largest images. I still don't know Netlify is so slow to build compared to my local computer, so we should look into that. I think this is the best I can do for now, and I just triggered a build in my branch to see the final results. I can see that other customers have asked Netlify to bump the build execution time, so I'm thinking that we should ask them about that too. Anyway, if this PR succeeds in building, should I merge it to `main`?

runemadsen:
@kfahn22 I did a quick resize to 1000px and a PNG optimization on the 20 largest images, but we should go through it more thoroughly and make sure that all contribution images are their smallest possible byte size. I'm not sure how small we want to go in the resizing?

kfahn22:
If you let me know exactly what size you want, I will resize the ones I have.

runemadsen:
@kfahn22 I think that 1000 pixels on the largest side is good for now. If we go lower, it'll be hard to show the images larger if needed at some point. It also helps running them through [a PNG optimization algorithm](https://tinypng.com/). Thanks!

kfahn22:
Sorry, I am thinking you mean 100X100.  Correct?

shiffman:
Hey! Yes, merge away! I'll be at venue within the hour and can poke around a bit more closely and merge some of the other pull requests if I can!

runemadsen:
Nice! The build is still on the threshold of timing out, so remember that it takes about 30 minutes for the build to complete when something is merged.

runemadsen:
Okay, it has been merged and the site is deploying. Let's hope it works :) I'm moving to the PCD venue.

PR #269, opened 2022-06-14
Images

kfahn22:
These are the 3 images that I found that were greater than 1 MB.  I ended up grabbing new screen shots with a  smaller  window size  because simply changing image dimensions (first commit) resulted in a fuzzy image.  It might be better to try the compression route with Squoosh.


runemadsen:
Very nice!

Issue #99, opened 2022-04-14
Track Thumbnails


shiffman:
Jason and I have been working on the track thumbnails. We have one ready to go for Nature of Code!

![image](https://user-images.githubusercontent.com/191758/163444302-b13737c4-f3c5-4e5e-ae38-6af181dc32e6.png)

JPG and source InDesign files:
https://drive.google.com/drive/folders/1v9JZhrC7ORPkwzpjAUWKo_zl_CG8wUvk

@fdoflorenzano Do you want to try incorporating this one and see if it works? Then we can produce the rest? Is this the right aspect ratio / resolution?

shiffman:
I want to open an issue to keep a list of what thumbnails on the site are still placeholder and what needs something new, I think this this one is already a long thread, maybe I'll close this and start a new one? @jasontheillustrator

PR #282, opened 2022-06-22
[WIP] First pass at trying new track thumbnails

shiffman:
This is what it looks like with me just selecting some stills from the image (similar to a YT thumb w/o text). Discussion at #99 

Screen Shot 2022-06-22 at 9 12 21 AM

Screen Shot 2022-06-22 at 9 12 32 AM

Screen Shot 2022-06-22 at 9 12 39 AM

Screen Shot 2022-06-22 at 9 12 46 AM


runemadsen:
Nice! I agree, I think this is the way to go. My suggestion would be to keep the same amount of slices in the same positions for all the track thumbs, as that will make everything seem a bit more aligned. If we make a small Figma template, it would just be a matter of clicking "Replace image" on each of the images to create a new thumb. Does your team use Figma?

shiffman:
> Nice! I agree, I think this is the way to go. My suggestion would be to keep the same amount of slices in the same positions for all the track thumbs, as that will make everything seem a bit more aligned. If we make a small Figma template, it would just be a matter of clicking "Replace image" on each of the images to create a new thumb. Does your team use Figma?

We don't but happy to try!

runemadsen:
I really like this @jasontheillustrator! I think this is the way to go. @shiffman What do you think?

shiffman:
I've strangely become very attached to the original slices and my variation but everyone I've shown this to prefers the single me in front of a set of slices! So I am good to go with that! @jasontheillustrator I know we've got a lot going on with animations for the WFC video, I think I'd prefer you be able to focus on that right now, can you share design files for the ones you made and I'll start integrating into the website and make some more? Then once WFC video is done, I can bring you back in to make tweaks, improve them?

jasontheillustrator:
What would be great is to have you @shiffman spend some time taking selfies specifically for these thumbnails. You could do a few wardrobe changes and a bunch of different poses. Maybe you bring back the train conductor outfit too. If you wanted to take it a step further, you could pretend to interact with some of the CT characters and I could illustrate them into the thumbnail. For wardrobe, it would be good to have a set with light colored clothes and a set with dark colored clothes so that there are options to choose from that could help you pop off the background screenshots.

jasontheillustrator:
Also, I made a thumbnail template and put it into a Drive folder:
Screen Shot 2022-06-23 at 11 00 30 AM

https://drive.google.com/open?id=13VEHulkCVFV1qQswEUuIpxAIs0kDGNrD

jasontheillustrator:
Just in case anyone needs a little help understanding the InDesign file, I left a bit of text in the template file explaining how to change out the images:
Screen Shot 2022-06-23 at 10 50 05 AM

shiffman:
Merging this so I have something to screenshot the website for the next video! @runemadsen @jasontheillustrator we'll continue to discuss next week!

Issue #244, opened 2022-06-11
No "Source Code" Link for Passenger Showcase


dipamsen:
For any contribution, the image and title links to the actual project (`url`), while the name of the creator links to their page (`author.url`)

![image](https://user-images.githubusercontent.com/59444569/173191820-013178d5-5e5a-43fe-a547-653323beb77f.png)

From the website, there is no way to open the source code of the contribution (`source`). Is this intentional? Nonetheless, this makes the `source` field in the contribution pointless.

shiffman:
Ah, this is an oversight! The `source` field is technically optional but it would be nice to link to it if it's there. I don't know if it'll make it too busy with the purple "play" icon but we could have a source code icon in the bottom right or top right? @fdoflorenzano @runemadsen what do you think?

fdoflorenzano:
To be clear, it was intentional about how we originally proposed it. The way this part was thought is that there's multiple ways of adding a showcase reference: `url`, `videoId` or `source`. The priority to actualy use is `url`, then `videoId` , and finally `source`. So`source` is not really pointless on it's own, it's pointless adding it with other types of sources because they have preference. I guess they didn't notice this when writing the contribution guide.

I guess we can try adding more icons in case they're there, but with the video option too it can get tricky really fast.

kfahn22:
I had noticed this, but decided that if anyone wanted to look at the code for a (p5 sketch) contribution  it is easy enough to modify  the link from Fullscreen to Edit.

shiffman:
Ah, thank you for clarifying @fdoflorenzano, this was my confusion! 

I think we can actually simplify this and only have `url`. `videoId` is not necessary since if the shared media is a video that can just be linked via the `url` property. The video is not being displayed / embedded here so we don't need to treat it as a special case.

These are probably 95% of the cases:

url:
* link to p5 web editor (includes source)
* link to open processing, codepen, or some other online editor (includes source)
* link to blog post documentation (which could separately link to source)

There are some rare cases where the url might link to:

* video on YouTube
* Deployed website

So I suppose for these rare ones we could allow an additional "optional" `source` property, but I'm also fine with just removing it entirely and saying a showcase submission is only:

```json
{
  "title": "project name",
  "url": "project url",
  "author": {
    "name": "Your Name (or nickname!)",
    "url": "Link to more about you! (your personal site, GitHub, social media, etc.)"
  }
}
```
[...]

PR #285, opened 2022-06-22
Ported CC 071 Minesweeper

DenisovichDev:
Ported Coding Challenge 071: Minesweeper.

- Added timestamps (wasn't in the YouTube video)
- Fixed as many links as possible
- Removed the inactive projects

**Things to consider**
- There should probably be a way to open the source codes of the contributions from the website.
- Clicking on the contribution card always opens the links from the "url" property. In case the "url" property is missing, it doesn't open the YouTube video from the "videoId" property. 

I see that these things were discussed in #244


shiffman:
Thank you for contributing @DenisovichDev! Showcase projects that no longer are active don't need to be included. They can always be resubmitted and they are archived on the old website (https://github.com/CodingTrain/website-archive).

DenisovichDev:
Okay, I'll keep that in mind!

content/videos/challenges/71-minesweeper/index.json
```diff
@@ -0,0 +1,85 @@
...
+  ],
+  "codeExamples": [
+    {
+      "title": "Minesweeper",
+      "description": "A simple Minesweeper game",
+      "image": "minesweeper.png",
+      "urls": {
+        "p5": "https://editor.p5js.org/codingtrain/sketches/-mz-ePoqd",
+        "processing": "https://github.com/CodingTrain/Coding-Challenges/tree/main/071_minesweeper/Processing",
+        "other": "https://github.com/CodingTrain/Coding-Challenges/tree/main/071_minesweeper"
```
shiffman, line 39: Noting we can skip the "other" property if it's just the same as the Processing code. Other is really just for if it's not p5 or Processing.

content/videos/challenges/71-minesweeper/index.json
```diff
@@ -0,0 +1,85 @@
...
+          "icon": "🎥",
+          "title": "Prototypes in JavaScript",
+          "url": "https://www.youtube.com/watch?v=hS_WqkyUah8",
+          "description": "In this video, I examine the concept of \"Prototype\" in the JavaScript programming language."
+        },
+        {
+          "icon": "🎥",
+          "title": "Inheritance with Prototype in JS",
+          "url": "https://www.youtube.com/watch?v=CpmE5twq1h0",
+          "description": "You should go and watch my videos on ES6 classes instead, but if you choose to be here, I will discuss how inheritance works with prototypes in JS."
```
shiffman, line 80:
```suggestion
          "description": "You should probably watch my videos on ES6 classes instead, but if you insist on being here, I will discuss how inheritance works with prototypes in JS."
```

shiffman (commented):
Fantastic, wow, thank you @DenisovichDev! I've made a few small comments. Here are some other notes.

1. I think I am going to try to get away from having redundant text in the preview image for the video (this is the file `index.jpg`. I might actually suggest just replacing it with the `minesweeper.png` image that is for the code example.

2. We've decided that the `source` property in the contributions is no longer applicable. See #244. So those can be removed. It's up to the showcase submission to include a link to the source (if they want) on the url they submit themselves. It's ok to leave them there for archival purposes, but I just want to note that they aren't seen on the site itself.

Let me know if you have any questions, thank you so much for doing this!!!

DenisovichDev:
Thanks for all the guidance @shiffman! I have resolved all the issues and it's good to go. I'll keep the things you mentioned in mind for the next PRs.

shiffman:
Amazing, thank you!! 🎉🎉🎉🎉

PR #298, opened 2022-07-01
Marching squares

kfahn22:
Adding c5-marching-squares

You also want the Coding in the Cabana videos  added as a separate track, correct?  What about the showcases?  There are A LOT so I would think duplication is not a good idea because that means adding a ton of images, but what is your thought on this?

I wasn't able to figure out how to grab your index.pngs for these videos.  If you tell me where to find them, I will add.   Or I could add one of the example pngs for the time being.


shiffman:
> I wasn't able to figure out how to grab your index.pngs for these videos. If you tell me where to find them, I will add. Or I could add one of the example pngs for the time being.

@kfahn22 let's use one of the example PNGs as a start! I generally prefer these to the YT thumbnails with text anyway. Sometime I'll want to do a pass over all the thumbnails maybe.

dipamsen:
@kfahn22 I don't think there is any duplication required, tracks can be made from any videos (even other coding challenges), it is not mandatory to create a specific folder for a track. See [example](https://github.com/CodingTrain/thecodingtrain.com/blob/main/content/tracks/side-tracks/algorithmic-botany/index.json)

kfahn22:
I have added an index.png.  Also did a pull request with images for other Coding in the Cabana videos.

shiffman:
> @kfahn22 I don't think there is any duplication required, tracks can be made from any videos (even coding challenges or videos in other tracks), it is not mandatory to create a specific folder for a track. See [algorithmic-botany](https://github.com/CodingTrain/thecodingtrain.com/blob/main/content/tracks/side-tracks/algorithmic-botany/index.json), which uses challenge videos and doesn't have its own folder.

@dipamsen is right! The engine of the website only requires a single JSON file (and set of images) per video but the videos themselves can be organized in multiple tracks and side-tracks! The showcase files will show up no matter where the video is listed too!

So do do this you'd want to create a  folder and JSON file here:

https://github.com/CodingTrain/thecodingtrain.com/tree/main/content/tracks/side-tracks

and then list the videos by path in the JSON file. Take a look at the site maintenance guide for more and feel free to pull request changes or make suggestions to that document if there are ways to improve it!

PR #291, opened 2022-06-26
Ported CC 119 Binary to Decimal Conversion

DenisovichDev:
Added challenge binary to decimal conversion. Please check if you are okay with the `index.png` and the thumbnail.

PR #322, opened 2022-07-05
Tweaking submission form

runemadsen:
This PR fixes a few things around the submission form:

- [x] Adds language around JPG uploads
- [x] Changes "your website" and "your email" to optional fields
- [x] Updates the Netlify function to handle optional fields

@shiffman I tested this twice and everything seems to be in order, but please give it a check before merging :)

### DIFF
--- a/netlify/functions/submission-background.js
+++ b/netlify/functions/submission-background.js
@@ -27,16 +27,21 @@ exports.handler = async function (event) {
   const imagePath = `${showcasePath}/contribution-${unix}.${postInfo.imageExt}`;
   const octokit = new Octokit({ auth: process.env.GITHUB_TOKEN });
 
-  // Get the SHA of the main branch
+  /**
+    Get the SHA of the main branch
+  **/
   const shaRes = await octokit.request(
     `GET /repos/${owner}/${repo}/git/ref/heads/main`
   );
   const mainSha = shaRes.data.object.sha;
 
-  // Make a new branch named after the user
+  /**
+    Make a new branch
+  **/
   const branchName = slugify(
     `showcase-${slugify(postInfo.authorName)}-${unix}`.toLowerCase()
   );
+
   const branchRes = await octokit.request(
     `POST /repos/${owner}/${repo}/git/refs`,
     {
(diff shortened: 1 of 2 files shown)


runemadsen:
I spent some time today debugging this, and I think this has something to do with Netlify. It's all working on my local machine with the same image, but once it's deployed, I get a `net::ERR_ABORTED` HTTP error with status `500`. Furthermore, our background function is never called, since no logging is happening. [We are not the first people to experience this](https://answers.netlify.com/t/netlify-function-returns-500-only-in-production-and-no-logs/13322/12) and none of the solutions offered in the thread is helping (I tried upgrading the Lambda to Node 16, etc).

@shiffman Should I send a support email or can we get in touch with your contact at Netlify? This seems like a problem on their end.

shiffman:
I'm going to go ahead and merge this and will submit a support request to Netlify! @jlengstorf tagging you here too in case you want to follow. 👋 Thanks for all your help and encouragement with the website!

(Also noting that this might be related to the netlify build issues I'm having with #291.)

jlengstorf:
Hey, @shiffman — I just ran a test (#357) with no email/website and it worked as expected. Are you still seeing this error?

shiffman:
Thank you for checking @jlengstorf!! I'm able to produce the error still. I emailed support and they suggested changing to Node 14.x but that didn't seem to resolve it. Here's the message I wrote to support.

> We currently are having a problem with the submission-background function in that everything works perfectly on localhost using netlify-cli, but on the Branch Deploy we get a net::ERR_ABORTED 500 error. Furthermore, our function is not called at all as it's not producing any logs. We poked around and saw that other people had similar issues (https://answers.netlify.com/t/netlify-function-returns-500-only-in-production-and-no-logs/13322/12), but we've tried all the proposed solutions and none of them work. So I think we need someone from Netlify to look at logs from your end to see where the error is.

I'm still able to get this error with this test image: https://drive.google.com/file/d/106hFcqve0iIc6J1qaByk3o8eZhz-so0y/view?usp=sharing

Screen Shot 2022-07-10 at 9 57 54 AM
Screen Shot 2022-07-10 at 9 58 05 AM

The logs oddly don't show anything:

Screen Shot 2022-07-10 at 10 04 13 AM

@runemadsen let us know if there are any updates on your end! It's not a critical error b/c the form seems to *mostly* work! Not sure why certain images trigger the error.

runemadsen:
@jlengstorf I was actually thinking about this too. The odd thing is that it works on localhost, where the Netlify functions works fine. It's breaking in production before the `POST` request reached the function as we have `console.log` first thing in the function and that is never logged. But perhaps it is what you say and an error that only happens in a specific Node version which runs in production or something like that. I would love to get to the bottom of this, but it's hard when we can't use the production logs.

jlengstorf:
I'm wondering if the `multipart/form-data` encoding is doing something that is incompatible with Netlify? If the same image works after re-encoding I feel like that'll be a huge clue we can dig into.

PR #325, opened 2022-07-05
Passenger showcase contribution from VBalko

ch00ooch00oo:
Yay!


shiffman:
Great work! I'm open to accepting some/all of these improvements as a pull request to https://github.com/CodingTrain/Wave-Function-Collapse if you want to open an issue there to discuss!

PR #332, opened 2022-07-06
Index images

kfahn22:
I have replaced the index images from the challenges I added. (I also found a small typo in the json file for Lévy flight and fixed.)  Would you like new images for index images for 53/52/36/25?  I started on that and then decided I would ask first.


shiffman:
Thank you @kfahn22!

PR #339, opened 2022-07-06
Passenger showcase contribution from darkfrei

ch00ooch00oo:
Yay!


content/videos/challenges/172-directional-boring/showcase/contribution-1657144189.json
```diff
@@ -0,0 +1,7 @@
+{
+  "title": "Horizontal Directional Drilling",
```
shiffman, line 2:
```suggestion
  "title": "Horizontal Directional Drilling in Lua",
```

shiffman (commented):
Is this accurate? (Adding "in Lua" to the title?) I think this will help distinguish it on the page!

shiffman:
I went ahead and added "Lua" to the title since I think it's an important note for the showcase!

PR #340, opened 2022-07-07
Fixes

kfahn22:
Hopefully these images will be better centered.  Also fixing link to repo in brownian tree (sorry about that !)


shiffman:
Thank you thank you!

PR #349, opened 2022-07-08
Ported CC 120 Bit Shifting

DenisovichDev:
Sorry this took a while to finish. Got a bit busy and this PR slipped my mind.

Things to mention:
 - You might wanna change the `index.png`, as it is not as appealing or colorful


klinegareth (commented):
@DenisovichDev any reason to keep the `contribution2.json.bad`? Otherwise this all looks good to merge!

DenisovichDev:
It was there just in case the contributors came to revise the links, but I guess now they can just do a new contribution. I'll delete it, no problem.

PR #350, opened 2022-07-08
Passenger showcase contribution from Sergey Alyaev

ch00ooch00oo:
Yay!


alin256:
I do not understand if or how it works? Was it added to theCodingTrain.Com ?

runemadsen:
Yup, we run an automated script that does two Git commits with your email based on the form upload. The script is here: https://github.com/CodingTrain/thecodingtrain.com/blob/main/netlify/functions/submission-background.js

alin256:
I am confused because this seems to be merged but I do not see it on the website.

shiffman:
I just checked and there was an error in the latest build, looking into it now.

shiffman:
An error slipped into main somehow, I've fixed it in #356.  Once that is merged and the site rebuilds it should show up! It's a slow process b/c the build time at present on netlify is about 30 minutes!

@runemadsen do the tests run also against pull requests or just the netlify preview build?

shiffman:
@alin256 FYI this is showing up now on the site!

alin256:
@shiffman, I do not see it. Am I looking in the right place: https://thecodingtrain.com/challenges/172-directional-boring ?

Screenshot 2022-07-11 at 10 23 03

shiffman:
Ok, this is indeed very strange! I should have taken a screenshot it b/c it was definitely there at one point! 

@runemadsen do you have any idea why this might be happening? When I'm building the site locally this has happened to me and an `npm run clean` resolves it. I'm hesitant to add `npm run clean` to each and every netlify build given the time that would add? 

I'll note I've only ever seen this for showcase submissions. Is it possible it doesn't rebuild the challenge page if a showcase project is added but no changes are made to the the actual video JSON file itself? Really odd that it would be there at one point and then disappear in a later build?!?

I can force a "clear cache and deploy main" but curious if you have any thoughts or want to run any tests first?

runemadsen:
That is indeed strange. Let me poke around from my end and see if I can duplicate this!

PR #360, opened 2022-07-10
Add tooltip to Contribution header for long names

dipamsen:
In many Passenger Showcase contributions, the names are long and are cut off.
image

This PR adds a tooltip which shows the full name of the contribution on hovering on its header. This is only applied to names longer than 30 characters long.

image
(Here, I am hovering on the contribution header. The mouse is not visible in screenshots.)


runemadsen:
Amazing. Thanks!

PR #363, opened 2022-07-11
Adding 54.2-islamic-star-law-of-sines

kfahn22:
Adding part 2 of challenge and updating link to part 2 in json of 54.1.  Noting that the contributions were the same as for part one so the showcase folder is a copy.


shiffman:
So the old site had a system for multi-part coding challenges but I've now moved to just releasing longer videos with chapters. I'm not sure the best way to handle these for the older ones. I think let's just put the contributions in the last part? Maybe then we put `canContribute: false` in the first part? And a note in the description that says contributions are in part 2? And make sure all the parts are linked via "related challenges" or a "group link"? I think this is better than duplicating them.

klinegareth:
@shiffman I think it actually makes sense to just duplicate the contributions, unless there are ones that specifically refer to a certain part of the coding challenge?

shiffman:
> @shiffman I think it actually makes sense to just duplicate the contributions, unless there are ones that specifically refer to a certain part of the coding challenge?

The reason I don't think this makes sense is b/c anyone submitting a future one would have to then know to submit it multiple times to multiple pages. And the site build is already so long this is just going to unnecessarily increase it? 

Here's a new idea. For older multi-part challenges just have one page for them that's associated with the first (last?) video. Then create a `grouplink` with the other parts. We could also make a side track for some of the more popular ones. 

I think we can bandaid this b/c I'm not doing multi-part challenges anymore and there aren't that many.

shiffman:
Thanks @klinegareth if the combined versions look good to you let's merge those and close these older ones!

kfahn22:
I am closing this one and creating two more.  One to delete 54.1 and a second to add the combined challenge.

PR #366, opened 2022-07-12
Passenger showcase contribution from Kathy McGuiness

ch00ooch00oo:
Yay!


shiffman:
Thank you @kfahn22! @runemadsen I noticed this one doesn't show up in the deploy preview which is probably related to our discussion in #350. I think I'll open an issue for us to track and do some investigating with a local build. @kfahn22 I'm going to merge this now, let's remember to check the website in about an hour to see if it shows up!

Issue #364, opened 2022-07-11
Code! track examples


shiffman:
Filing this issue as a reminder to add links to the p5 web editor versions of the code demonstrated in the Code! track. I'd like to make sure this is done by start of school in September. Not sure if these are all easily findable in the `codingtrain` p5 web editor account or if I need to re-create these.

dipamsen:
I would like to work on this! I am starting by scraping the examples data from the old website.

shiffman:
@dipamsen can this be closed or is there more work to be done? Thank you for all the work you put into this!!!!

dipamsen:
This can be closed!

shiffman:
Ok, I'll open a new issue to track this list!

PR #368, opened 2022-07-13
Add data for Code! from old website

dipamsen:
See #364 .

This PR ports all the data from the old website for the Code! Track.
- p5.js Web Editor Code Examples have been added, without thumbnail images
- Timestamps have been added for the videos which had timestamps in the old website.
- `groupLinks` (Links and Videos) have been also ported from the old website.
- All passenger showcase submissions have been added, without the `source` field.

Other changes: 
- ~~Removed `videoNumber` property and added video number to the title (Same as NOC track)~~
- Renamed some files to match video numbering in the videos.
- Removed duplicate video
  image
  Here, both the files had the same video - Mouse Interaction, but different numbering. The directories were called `4-for-of` and `5-mouse-interaction`. I have deleted the `4-for-of` folder. Was this supposed to be a different video in this track?

Note: All the JSON files in this PR have been [programatically generated.](https://github.com/dipamsen/CodingTrain-Scrape)


shiffman:
> Removed videoNumber property and added video number to the title (Same as NOC track)

This is an interesting question! @klinegareth @runemadsen what do you think about us incorporating the number field into the track video themselves? @dipamsen I think the reason we don't want to have it in the title is that videos can be remixed into different side tracks and then the number might add confusion. I'm actually fine with not having any numbers in the tracks at all, I think the interface does a good job of showing the chapters and order without them? We can remove them from the Nature of Code track too depending on what we decide.

shiffman:
> Here, both the files had the same video - Mouse Interaction, but different numbering. The directories were called 4-for-of and 5-mouse-interaction. I have deleted the 4-for-of folder. Was this supposed to be a different video in this track?

@dipamsen I honestly don't remember, I think we should keep the folder names and concise as possible. The directories are just for us to keep things organized so feel free to organize the files as you see fit!

dipamsen:
> I think the reason we don't want to have it in the title is that videos can be remixed into different side tracks and then the number might add confusion.
I have reverted back this change. 

Even if video numbers are incorporated in the website, the issue still remains. When these videos will appear in other side tracks, their video numbers will be the same as that from the main track.
I think we should keep these as is and not include the video number.

klinegareth:
@shiffman @dipamsen I think it makes sense to keep the video titles as-is (without numbering), especially since they're already organized into track chapters. The directory numbering is from before we decided to simplify video urls, so it was mainly a way for me to keep track of what videos had been added, belonged to which chapter, etc. I think those could definitely be removed, unless there's some other reason for keeping them.

shiffman:
Sounds good to me! If we have the data we might as well store the number under a `videoNumber` field in case we ever change our mind and want to use it. . .

klinegareth:
Looks good @shiffman ! Merging now

PR #380, opened 2022-07-16
- Adding showcase contribution

jskrist:
This contribution shows a solution for dynamically determining edge conditions for matching tiles, so the user can select what ever folder contains images to be tiled, and the code automatically assigns edge labels and determines "fuzzy" matches.


shiffman:
Wow, so cool, thank you!

PR #398, opened 2022-07-25
Challenge index thumbnails 

shiffman:
Working through the challenges to create thumbnail images that do not include title text or photo of me (unless specifically relevant for the visual itself). This is what the first page looks like. I should probably redo the mandelbulb one, the point cloud is too subtle.

See preview: https://deploy-preview-398--codingtrain.netlify.app/challenges

Screen Shot 2022-07-25 at 1 01 51 PM

@kfahn22 what do you think of these so far?


dipamsen:
These look great! I was thinking of doing something like this for the Cabana videos - to only showcase the code example output from the thumbnail:

0ZONMNUKTfU-HD

shiffman:
> These look great! I was thinking of doing something like this for the Cabana videos - to only showcase the code example output from the thumbnail:

Yes, I love this, feel free to pull request these! You are also welcome to make any changes, play with color (see: https://coding-train-colors.vercel.app/). I believe the aspect ratio should be 16:9 for these also FYI.

Here's another hack, if you go to https://editor.p5js.org/codingtrain/sketches you can search for "thumb." If I remember it correctly I often name the sketch I use to make the thumbnail with the keyword "thumb". Enjoy!

shiffman:
I'm going to go ahead and merge this, any other challenge images can be in a separate branch. @dipamsen and @kfahn22 feel free to coordinate so that you don't duplicate any efforts!

PR #401, opened 2022-07-25
New topic select

bravomartin:
This PR replaces the topic filter with a new top selector with search functionality, making it easier to navigate through the topics.

demo


shiffman:
ooooh, this is exciting!

runemadsen (changes requested):
This is excellent @bravomartin! I just fixed the build, so the preview is now working. I think we also need to change the language selector, so we have the two selectors side by side. If you can do that, I vote for merge afterwards!

bravomartin:
Added the filter-select to languages. please take another look and approve / merge if it works for you

runemadsen:
This looks great to me! @shiffman feel free to merge if you don't have any input. As far as the lowercase/uppercase problem, I do agree that a `Processing` and `process` topic should lead to the same filter. I think we could fix this by adding notes to the content tests or just lowercasing things in the URL.

shiffman (approved):
Woohoo!

shiffman:
> This looks great to me! @shiffman feel free to merge if you don't have any input. As far as the lowercase/uppercase problem, I do agree that a `Processing` and `process` topic should lead to the same filter. I think we could fix this by adding notes to the content tests or just lowercasing things in the URL.

Would it be crazy to do both? Merging!

PR #399, opened 2022-07-25
Add canContribute property to NOC track

dipamsen:
Fixes #390 

- Added canContribute property to videos
- Removed extra showcase
- Added code examples and references to Perlin Noise section

(Due to some reason, I forgot to create a branch and commited to main directly.)

PR #402, opened 2022-07-25
Tests expect videoNumber string in JSON

shiffman:
I noticed the tests failing for #401 and #399. This should fix the issue which was floating point numbers appearing for the `videoNumber` property in the video JSON.

Unfortunately, with my global find / replace I also changed a lot of whitespace / indentation so looking at the diff is a bit of a mess.

I also fixed an error where some "showcase" folders had the old name "contribution." 

@dipamsen let me know if this ok in terms of the change I made to the Code and Data-API tracks!


dipamsen:
Looks good!

PR #403, opened 2022-07-26
Challenge images with Coding Train colors

kfahn22:
Adding a bunch of index images.  I see that I forgot to edit size of a couple of the images down to 800px.  Can you do this or should I redo?


shiffman:
These are fantastic! Thank you @kfahn22! A couple notes for future reference (but no need to change unless you want to.)

* I think it's ok to use white as the foreground stroke color in addition to the CT colors if you like.
* You can sometimes increase the `strokeWeight()` from whatever I had in the example to create a bolder, higher-contrast image.
* Let's do 1280x720 for these! There are a limited number unlike the showcase submissions.

PR #410, opened 2022-07-28
Challenge images 

kfahn22:
Here are some new images.  I am (I think) changed dimensions to 1280 by approx 720.   Also changed stroke to white for a lot of them.  What do you think of these?


shiffman:
These are fantastic!

PR #412, opened 2022-07-30
New images

kfahn22:
Here are some new images.  I have added some color gradients to some of the ones that originally had black backgrounds to add a little bit of color, but keep original aesthetic.  Also  increased strokeWeight on some of them.


shiffman:
This is incredible!

Issue #367, opened 2022-07-12
Missing Showcase Projects


shiffman:
I'm opening an issue to track the discussion in #366 and #350. Some passenger showcase projects do not appear on the website after a successful build. My working theory is that there is some caching issue in the build process. I believe an `npm run clean` before `npm run build` resolves the issue. However, I'm not sure it makes sense to add `npm run clean` to each and every netlify deployment. 

@runemadsen I'll see what I can investigate this week and we can return to looking at this next week when you are available.

shiffman:
Noting that last night when no other builds were running I ran a force "clear cache and deploy site."

<img width="536" alt="Screen Shot 2022-07-13 at 9 01 03 AM" src="https://user-images.githubusercontent.com/191758/178739645-87fb5221-72c1-447d-8f9d-f7add23d83d8.png">

The build took a ~67 minutes, but the missing showcase projects do now appear on the site. I may be mistaken but we should keep a close eye on these pages as I believe when this occurred before later deployments caused the projects go go missing again. We'll see if that happens this time.

<img width="323" alt="Screen Shot 2022-07-13 at 8 58 55 AM" src="https://user-images.githubusercontent.com/191758/178739254-53664c55-b9ac-46d4-976c-abc947087faa.png">

Tagging @kfahn22 and @alin256 in case you want to double-check your submissions.

alin256:
The submissions showed up now, but I sent in 2 due to confusion. I can PR later to remove one.

@shiffman, I wonder if 172 should show up on the front page as the latest challenge? 
I did not figure out the gatsby layout, and I do not know if it is desired behaviour.

dipamsen:
@alin256 the challenges featured on the home page are not dynamic, they need to be manually updated in [this file](https://github.com/CodingTrain/thecodingtrain.com/blob/main/content/pages/homepage/index.json)

shiffman:
Unfortunately, after merging pull request #369 the showcase projects referenced here are missing again ([direct link to deployment in case its helpful](https://app.netlify.com/teams/shiffman/builds/62d0152ab9d3cd0008ec6d2b)). I'm going to run `clear cache and retry deploy`. I suspect there is something going on with the cache that persists even across later builds where the cache isn't cleared? Is that possible? [new build for tracking purposes](https://app.netlify.com/sites/codingtrain/deploys/62d0cd763e0408341b0180b4).

shiffman:
[...]

PR #414, opened 2022-08-01
Removed a duplicate show-case that I have submitted earlier

alin256:
The double-submission was caused by Issue #367 which is now resolved.


shiffman:
Thank you @alin256!

PR #424, opened 2022-08-05
New images

kfahn22:
Adding more index images.  Also changing link for the contribution in shape classifier to source.


shiffman:
Beautiful work!!

PR #425, opened 2022-08-05
Re-Ordering the Sections on the Getting Started Guide

### LINKED ISSUE #409: re-order "getting started" guide
Maybe beginner p5 needs to be at the top?

shiffman:
This will close #409. I've updated the guide to lead with the beginner tracks.


dipamsen:
Not sure if this is required, but I believe in a previous livestream you mentioned that you preferred the header text ("Code! Programming with p5.js") to link to the track page.
image

shiffman:
Ah, yes, I remember noticing this! I just tried to add it but I get an error with any of the following . . . any ideas? @fdoflorenzano do you happen to know what might be preventing me from making the header a link?

```md
### [Code! Programming with p5.js](/tracks/code-programming-with-p5-js)
```

```md
### <a href="/tracks/code-programming-with-p5-js">Code! Programming with p5.js</a>
```

```html
<h3><a href="/tracks/code-programming-with-p5-js">Code! Programming with p5.js</a></h3>
```

Screen Shot 2022-08-05 at 3 10 55 PM

dipamsen:
The `kebabCase` function is being called with the contents of the header tag. 

(This function is responsible for creating ids, i.e. `code-programming-with-p5js` which is used in the table of contents for linking to a section, [like this](https://thecodingtrain.com/guides/getting-started#code-programming-with-p5js))

image

The function is having trouble preparing an id for this section (`code-programming with-p5js`) because, it expects a string as its argument. However, it recieves the contents of the header, which happens to be the `<a>` tag. 

If we remove the `kebabCase` call (line 65 above), suprisingly everything works fine. This is because, for Level 3 headings, the Table of Contents does not create an entry, so it technically does not require an ID field. 

image

However, I do not know if this is an actual solution to the issue.

shiffman:
Hmmm, maybe I'll revisit this later, I also don't like how it looks with the color and underline. Perhaps it's not critical. @klinegareth if you want to take a quick look at this to make sure it's all good then merge go ahead! I can open a new separate issue to think about linking some of the titles (I had the same thought in a few other places on the site).

Issue #434, opened 2022-08-10
Showcase cover images do not work in tracks


dipamsen:
Cover images for showcase submissions do not appear on the website. It uses the placeholder image, even though an image with the same name as the json file is defined.

<details>
<summary>Examples</summary>

* [NOC 3.1 - Angles and Rotation](https://thecodingtrain.com/tracks/the-nature-of-code-2/noc/3-angles/1-angles-and-rotation) 
![image](https://user-images.githubusercontent.com/59444569/183851385-6200b95d-1ac6-43b2-afdc-b6b2aa9f5af4.png)
![image](https://user-images.githubusercontent.com/59444569/183851543-6351351f-1e62-484f-a5ef-d222f8e5e473.png)

* [Data and APIs 3.5 - Web Application Deployment](https://thecodingtrain.com/tracks/data-and-apis-in-javascript/data/3-the-weather-here/5-web-application-deployment)
![image](https://user-images.githubusercontent.com/59444569/183852138-c5ef55c0-1fce-40f9-9ad0-2db181b21654.png)

![image](https://user-images.githubusercontent.com/59444569/183852017-51c3fb70-8bd8-4506-aadf-9cd2612e3776.png)

</details>

This behaviour is observed in all tracks (side/main), except for those side tracks which are created from videos in `challenges` (algorithmic botany, coding in the cabana). It seems like the showcase cover images for videos outside `challenges` are not being loaded properly.

shiffman:
Ah, thank @dipamsen! I didn't realize this was more than just the one 3.1 NOC video. @klinegareth and I will look into this. @fdoflorenzano if you have any ideas let us know!

shiffman:
Thank you @fdoflorenzano! @kfahn22 after the site builds, go ahead and check video 3.1, we can re-open if there are still issues but should be working now!

shiffman:
@kfahn22 alerted me to this particular showcase image which is not showing up (although 3.1 is indeed fixed!):

* https://github.com/CodingTrain/thecodingtrain.com/blob/main/content/videos/noc/3-angles/4-polar-coordinates/showcase/contribution3.png
* https://thecodingtrain.com/tracks/the-nature-of-code-2/noc/3-angles/4-polar-coordinates
 
[...]

PR #442, opened 2022-08-11
fix: update code example cover image linking

fdoflorenzano:
In response to #434

### DIFF
--- a/node-scripts/node-generation.js
+++ b/node-scripts/node-generation.js
@@ -107,9 +107,7 @@ exports.createVideoRelatedNode = (
       codeExamples: (data.codeExamples ?? []).map((example) => ({
         ...example,
         image: createNodeId(
-          `cover-image/${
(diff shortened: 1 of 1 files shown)


fdoflorenzano:
@shiffman it isn't working here either (on the preview)! Maybe try clearing up the cache?

fdoflorenzano:
@shiffman I was able to re-build restarting cache for this branch on netlify and everything showed up. 

I'm guessing this is needed because the recent changes changed how Gatsby defined some data nodes, and it also relies a lot on caching those nodes. So maybe it update how a lot of these nodes relate to one another. Hopefully a re-build on main also does the trick.

shiffman:
Thanks @fdoflorenzano I'm going to merge this and then will do a force rebuild and clear cache! Hopefully that resolves, will keep you posted!

PR #440, opened 2022-08-11
154: Tic Tac Toe Minimax - Add Timestamps and Showcase

dipamsen:
Closes #343 

By mistake I have branched directly from the space invaders branch, so this should be merged after #435

PR #446, opened 2022-08-13
Fixed duplicate allcontributors entry

klinegareth:
Fixed an accidental duplicate contributor left in from merging #440


shiffman:
Thanks for catching this!

PR #452, opened 2022-08-15
Ported CC 12 Lorenz Attractor

DenisovichDev:
This PR ports CC 12 Lorenz Attractor

Comments:
- All the contributions are still available
- For some reason the development build wasn't working, so I deleted `node_modules` and did `npm install` and that seemed to have fixed it
- Not sure why the netlify deploy preview isn't showing the contributions, because it works fine locally


DenisovichDev:
Hmm. For some reason my local builds are failing. Might need an npm install.

shiffman:
There seems to be an issue where contributions don't show up unless I "clear cache" and build. I'm merging a bunch of things now and then will trigger the force update, let me know if they still don't show up!

PR #456, opened 2022-08-17
Fix zero matches

### LINKED ISSUE #397: Showing "0 matches found" even when filter is off
image
Shows message "0 videos within this track match the tag:" when no topics/languages are selected.

runemadsen:
This PR fixes #397 while optimizing the filtered path rendering a bit. 

@shiffman Feel free to merge if the preview looks good.


shiffman:
This is looking great! I think the "matching" could use a little refining? Maybe a "." instead of a ":" or list the matched terms is maybe better?

> 3 videos within this track match the tag _processing_.

Also should "tag" say "language" or "topic"? Or should we just go ahead and change "topic" to "tag"?

Finally, there is maybe an extra line / column when there are more than 2 videos. Or is this the correct behavior?

https://deploy-preview-456--codingtrain.netlify.app/tracks/lang/processing/topic/all

topics

runemadsen:
Great! How about this:

> 3 videos within this track match the filter

I'll take a look at the box weirdness.

shiffman:
> Great! How about this:
> 
> > 3 videos within this track match the filter
> 

Works for me!

shiffman:
Apologies @runemadsen I think a merge conflict popped up here. I'm happy to work to resolve it but thought you might prefer to take a look.

fdoflorenzano:
> I'll take a look at the box weirdness.

Hey! I'm just wondering what's weird about it. I'm under the impression it's the expected result based on out designs. What would you expect it to look like?

runemadsen:
Ah, I see. I think it's because when there are odd number of tracks, the bottom right slot doesn't have a small box to separate the two slots. But when it's filled, it looks nice and mirrored:

Screen Shot 2022-08-23 at 8 57 31 AM

I'm fine leaving this as is, but we could also add a small box when there a odd number of results. @shiffman thoughts?

fdoflorenzano:
I added the missing square! https://deploy-preview-456--codingtrain.netlify.app/tracks/lang/leaflet-js/topic/all

Screen Shot 2022-08-23 at 10 34 41

shiffman:
Looks amazing, thank you so much @fdoflorenzano and @runemadsen!

Issue #205, opened 2022-06-09
Port YouTube Description Generator Script


shiffman:
The old site had [a super convenient script to take a markdown page for a video and generate the YouTube description](https://github.com/CodingTrain/website-archive/blob/main/_scripts/generate-youtube-descriptions.js) for it. It was originally written by the amazing @duskvirkus!

I'd like to eventually port that script to work with the new site and (dare I say) maybe even eventually integrate with the YouTube API itself???

I know @klinegareth has been thinking about this and developing a CLI for the site as well. Let's discuss!

I was reminded of this when thinking about #203 and redirects and all the links from YouTube video descriptions that are now broken. 😬

dipamsen:
Here is a draft version I created: https://gist.github.com/dipamsen/8100b65ec03262a38cf066cd6e67c837

shiffman:
This is fantastic @dipamsen! I don't have time to review it right now but I will take a look very soon! I don't know the best way to integrate additional scripts but I could imagine something like the following being very useful. 

```console
npm run yt-desc /path/to/video
```

Or maybe we even make it an endpoint? @fdoflorenzano @runemadsen would adding a YouTube description generator script into this repo affect the website code / build process in any way that I might not be anticipating?

fdoflorenzano:
@shiffman Can you explain what you expect from this script? I'm not sure why a YouTube descriptions script needs to live in this repo. It wouldn't create content for the actual website, right?

runemadsen:
@shiffman @fdoflorenzano Jumping in here :) I remember us talking about this script earlier on, and as far as I remember, it's for simplifying the task of adding a video to the website and then to YouTube. So it just spits out the stuff needed for YouTube based on the website content. In that sense, I don't mind it being in this repo, and I think adding it as a script in `package.json` is a good idea. I propose adding it as a stand-alone script to `node-scripts`. It shouldn't affect the build process in any way. Happy to look through a PR if you tag me in it.

shiffman:
Thanks everyone! @dipamsen do you want to bring this over in a branch here so we can start collaborating on it? This is a nice opportunity to add some features and improve this script. Tagging @KobeLiesenborgs to follow as well!

dipamsen:
@shiffman Created a branch [here](https://github.com/CodingTrain/thecodingtrain.com/tree/description-script).

[...]

PR #461, opened 2022-08-19
Tweaks to YT Description script

dipamsen:
- Added property `playlistId` for main tracks to work with `yt-desc` script (see https://github.com/CodingTrain/thecodingtrain.com/issues/205#issuecomment-1159516945)
- Changes to script to use the `playlistId` for Previous Video / Next Video / All Videos
- Minor tweak to emoji selection for reference links
- Removed 404 amazon influencer store link from description (let me know if this needs to be kept)


shiffman:
Thank you for working on this more @dipamsen!!

shiffman:
> Removed 404 amazon influencer store link from description (let me know if this needs to be kept)

Yes, I no longer have an amazon affliate account. I'd like to create a https://bookshop.org list but I haven't had the time yet.

PR #468, opened 2022-08-21
Adding 104-Linear regression-with-tensorflow.js

kfahn22:
This is ready for review.  Please let me know if anything needs to be changed.

 Minor point -- I couldn't find the twitter icon (at least in the right format) for the reference to the tweet.


content/videos/challenges/104-linear-regression-with-tensorflowjs/index.json
```diff
@@ -0,0 +1,96 @@
...
+            "title": "References",
+            "links": [
+                {
+                    "icon": "💾",
+                    "title": "TensorFlow.js",
+                    "url": "https://js.tensorflow.org/",
+                    "description": "Webpage with reference materials and documentation about TensorFow.js"
+                },
+                {
+                    "icon": "📈",
```
shiffman, line 59:
```suggestion
                    "icon": "🐦",
```

We could use a bird!

Issue #417, opened 2022-08-03
Order of Tracks


shiffman:
I'm curious about how the order of tracks is set on https://thecodingtrain.com/tracks? Is it most recently updated? Random? By date?

I think I might like to manually set the order in a JSON file. What do you think @runemadsen?

runemadsen:
Good question. [This is the code we use to load tracks](https://github.com/CodingTrain/thecodingtrain.com/blob/main/src/templates/tracks.js#L83-L89) and there is no sort. So I'm guessing it's defaulting to the order in which the nodes were created? It's very easy to add a sort based on a field, and it can be date or just a number you can set to curate the order. Any preference?

shiffman:
I think to start let's add a "sort order" number so I can curate. I can always change it to something else later.

runemadsen:
Great, will do

kfahn22:
I think it would be useful for the tracks to be subdivided into MAIN and SIDE.  It isn't obvious when you look at the page that MAIN tracks are listed first and then SIDE tracks.   As for sorting, my vote would be alphabetical (after the featured track in each subcategory.)

dipamsen:
For ordering, I think it might be cleaner to have a `tracks.json` file somewhere listing the track slugs in order; and import it and sort the tracks accordingly?

shiffman:
This is ready to go! Open to any suggestions on the best order!

PR #481, opened 2022-08-23
feat: adds ability to sort tracks in track page

fdoflorenzano:
As proposed in #417, this adds the loading of an JSON file that defined the order in which tracks are shown in the Track Page. The file should be located in `content/tracks/index.json`, and should be an object with a `"trackOrder"` property that is an array of slug strings for the tracks. If a track is not found in this array, it will get assigned a very high number.

This also adds the front-end graphql query, and tests for this new file.

### DIFF
--- a/content-testing/content-structure.js
+++ b/content-testing/content-structure.js
@@ -1,6 +1,7 @@
 const {
   video: videoFormat,
   track: trackFormat,
+  trackOrder: trackOrderFormat,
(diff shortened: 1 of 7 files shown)


runemadsen (approved):
Beautiful! This is ready to go in my end.

shiffman (approved):
Amazing! Fantastic! Thank you!

PR #480, opened 2022-08-23
File content fixes

fdoflorenzano:
I noticed some fails when running content tests on main. This resolves the few issues found.

PR #482, opened 2022-08-23
Create node.js.yml to auto-run tests

fdoflorenzano:
This adds a GitHub Action workflow that runs the content tests on pull requests. This way, an extra layer of checking all files are correctly setup is added.

The PR fails cause there's errors currently on main. Which I propose to fix in #480


runemadsen (approved):
Amazing! This looks great to me. @shiffman feel free to merge.

fdoflorenzano:
Noting that these should be merged after #480, and merge main in here just to check that it correctly fixes the content tests.

shiffman:
Yes! I think I got this working!!

PR #490, opened 2022-08-25
Passenger showcase contribution from Talal 

ch00ooch00oo:
Yay!


shiffman:
Hi this is great! Was there a particular video you watched when creating it? Maybe we can find a better spot than the BASIC challenge since this is p5js!

shiffman:
Hi @Talalkanjo01! Just checking in one last time, is there a video on the site this showcase project is related to? Maybe you were watching one of the Nature of Code videos?

fturmel:
@shiffman I suggest we close this, no progress for months. They can always resubmit!

PR #500, opened 2022-08-28
Fix for coding challenge 70

shiffman:
@kfahn22 note that the property `github` is not valid. We can use `other` for anything that is not p5 or Processing. However, I think it's simpler just to link to the p5 editor sketch here when we have it.


kfahn22:
Sorry, I am not sure why I thought that was necessary.

PR #515, opened 2022-08-31
Fix minor UI problems on small screens

lobertrand:
I noticed 3 minor UI problems when viewing the website on my phone :
- On the Homepage, a horizontal scrollbar appears on a certain screen width, due to : 
  - (A) An overflow in the "CHALLENGES" `.subheader`
  - (B) An overflow in the footer `.navBox`
- (C) On the Tracks and Challenge parts timelines, the line which goes through seen videos appears empty on some screen densities or zoom level

I fixed these details using CSS.

(A) Before : a horizontal scrollbar appears and on the smallest size, the text is slightly off-centered 
a-before

(A) After : I made the "getting started" title wrap lines and fixed the center alignment on the smallest size
a-after

(B) Before
b-before

(B) After : I used an `overflow: hidden` to fix this (it was already applied but only on smaller screen sizes)
b-after

(C) Before (zoom level 175%)
image

(C) After (zoom level 175%)
image

### DIFF
--- a/src/components/Footer.module.css
+++ b/src/components/Footer.module.css
@@ -117,7 +117,7 @@
     font-weight: 900;
     font-size: var(--maru-medium);
     text-transform: uppercase;
-    line-height: var(--baseline);
+    padding: var(--box-padding) 0;
     margin: 0;
   }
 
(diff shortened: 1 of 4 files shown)


shiffman:
Great catch @lobertrand thank you for the fix! I haven't been looking at the site much on mobile so haven't noticed these. @fdoflorenzano if this looks good to you I will merge!

fdoflorenzano (approved):
Looks great! Some great catches! Thank you so much for this!

PR #521, opened 2022-09-02
New index images

kfahn22:
Mandelbulb -- not sure this is better?
Koch fractal snowflake higher strokewieght
Bit shifting color image


shiffman:
I like them!

PR #528, opened 2022-09-03
Running NPM Audit Fix

shiffman:
Let's see if the site builds properly still. . .


shiffman:
@runemadsen any reason I shouldn't be periodically updating these packages? (I ran `npm audit fix` here in this branch.)

runemadsen:
There's no problem updating minor version of packages with `npm audit fix`! I would recommend that if you update the `gatsby` packages, you update all of the official packages together since they are all synced by version. So this PR would need to update `gatsby-transformer-json` and `gatsby-transformer-sharp` too.

shiffman:
Thanks @runemadsen!

PR #537, opened 2022-09-03
YT Desc: Add Support for Multipart Challenges 

dipamsen:
* Parses multipart challenges and creates separate file for each part
* Changed output filename `{videoId}.txt`

Test:
```
npm run yt-desc challenges/35-traveling-salesperson
```

The filename change is required because otherwise all parts will be written into the same file. However, this might be a problem when this script is being used for new challenges for which `videoId` is not created yet. What should be the filename for the output files?

### DIFF
--- a/node-scripts/yt-description.js
+++ b/node-scripts/yt-description.js
@@ -90,7 +90,7 @@ function getVideoData(file) {
     );
   }
 
-  if (parsed.parts && !parsed.videoId) {
+  if (parsed.parts && parsed.parts.length > 0) {
     // Multipart Coding Challenge
     // https://github.com/CodingTrain/thecodingtrain.com/issues/420#issuecomment-1218529904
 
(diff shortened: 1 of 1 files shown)


node-scripts/yt-description.js
```diff
@@ -311,7 +311,7 @@ Music from Epidemic Sound
 
 This description was auto-generated. If you see a problem, please open an issue: https://github.com/CodingTrain/thecodingtrain.com/issues/new`;
 
-  let filename = data.videoId;
+  let filename = data.videoId || /\/((?:.(?!\/))+)$/.exec(pageURL)[1];
```
shiffman, line 314: I would prefer the following:

```
slug-name.txt (no videoID)
slug-name_videoID.txt (if videoID)
```

It's very hard to find and sort files by videoID.

shiffman (commented):
Wow! One comment re: filenaming.

shiffman:
Woohoo!

Issue #558, opened 2022-09-12
Fluid Simulation Page Port


shiffman:
In reviewing the site's logs I see a bunch of 404s for  `/CodingChallenges/132-fluid-simulation.html` so just noting as a good challenge to prioritize for porting! If and when we should also remember to add a redirect to [`redirect.json`](https://github.com/CodingTrain/thecodingtrain.com/blob/main/redirects.json). See #155 for tracking challenge ports.

shiffman:
I forgot @kfahn22 is already working on this one!

PR #559, opened 2022-09-12
Additional Redirects

shiffman:
Adding these after reviewing the site's logs. Only one missing is Fluid Simulation as referenced in #558


shiffman:
This along with #560 closes out #558

PR #565, opened 2022-09-12
Quad tree

kfahn22:
This is ready for review.


shiffman:
Hey! This is terrific! I would use an image that shows the lines of the quadtree itself as that represents the topic better than just the dots. Something along the lines of:

image

Let me know if you'd like to add it or if you prefer I merge and update the image afterwards.

kfahn22:
For the index image?

kfahn22:
quad
Is this OK?

shiffman:
Looks great!

Issue #262, opened 2022-06-13
Related Challenges


shiffman:
Noting that in #261 I removed "Related Challenges" (from tic-tac-toe for example) since the tests fail if we try to link to a challenge that is not on the site yet. When all the challenges have been ported, I'd welcome a pull request that does a pass over all of them to link up the related ones!

I'm forgetting, can we use the "related challenges" property in videos that are part of tracks also? @fdoflorenzano

fdoflorenzano:
@shiffman Yes! You can with all types of videos.

shiffman:
I believe this can be closed now that there are related challenges built out through all the content. @kfahn22 what do you think?

kfahn22:
I think there might still be a couple without related challenges, but I think this can be closed.

PR #572, opened 2022-09-16
Suggestion : Randomized related challenges view

lobertrand:
Hi! The "Try a Challenge!" section on a challenge page only shows the first 2 challenges from the "relatedChallenges" array. So putting more than 2 challenges in this array seems kind of odd. Unless we randomize the challenges that are shown on the page.

So this is my suggestion :smile:  I don't know if that's a good idea. One drawback of this is that this doesn't seem to work great with SSG (Static Site Generation).

Preview example : https://deploy-preview-572--codingtrain.netlify.app/challenges/162-self-avoiding-walk

### DIFF
--- a/src/components/ChallengesPanel.js
+++ b/src/components/ChallengesPanel.js
@@ -6,7 +6,7 @@ import Image from './Image';
 
 import * as css from './ChallengesPanel.module.css';
 import { getReadableDate } from '../hooks';
-import { shuffledCopy, useIsSSR } from '../utils';
+import { shuffleCopy, useSSR } from '../utils';
 
 const Card = ({
   className,
@@ -45,7 +45,9 @@ const Card = ({
                 imgClassName={css.image}
                 alt={`"${title}" challenge`}
               />
-            ) : null}
+            ) : (
+              <div className={css.image}></div>
+            )}
           </Link>
           <p className={css.date}>
             <span>
@@ -75,37 +77,34 @@ const Card = ({
 const ChallengesPanel = ({
   challenges,
   placeholderImage,
-  headerType = 'h2',
-  shuffle = false
+  headerType = 'h2'
 }) => {
   const Header = headerType;
-  const isSSR = useIsSSR();
-  const [suggestions, setSuggestions] = useState(
-    // This initial value is used on server side rendering
-    shuffledCopy(challenges, shuffle).slice(0, 2)
+  const { isServer } = useSSR();
+  const [suggestions, setSuggestions] = useState(() =>
+    // Empty placeholders on server side render
+    challenges.map(() => ({})).slice(0, 2)
   );
   useEffect(() => {
-    // This value is used on client side hydration, to shuffle the challenges
-    setSuggestions(shuffledCopy(challenges, shuffle).slice(0, 2));
-  }, [challenges, shuffle]);
+    // Shuffled challenges on client side hydration
+    setSuggestions(shuffleCopy(challenges).slice(0, 2));
+  }, [challenges]);
   return (
     <section className={css.root}>
       <div className={css.titleBox}>
         <Header>Try a challenge!</Header>
         <p>Suggested by the video you're watching</p>
       </div>
-      {/* This "key" attribute forces a fresh rerender on client side hydration,
-          which prevents a weird mix of static and dynamic content */}
-      <div className={css.challenges} key={isSSR}>
+      <div className={css.challenges}>
         {suggestions.map((challenge, index) => (
-          <Fragment key={challenge.videoId}>
+          <Fragment key={challenge.videoNumber}>
             <Card
               className={css.challenge}
               challenge={challenge}
-              placeholderImage={placeholderImage}
+              placeholderImage={isServer ? null : placeholderImage}
               headerType={`h${parseFloat(headerType[1]) + 1}`}
             />
-            {index !== suggestions.length - 1 && (
+            {(suggestions.length === 1 || index !== suggestions.length - 1) && (
               <div className={css.spacer}></div>
             )}
           </Fragment>
--- a/src/components/ChallengesPanel.js
+++ b/src/components/ChallengesPanel.js
@@ -6,7 +6,7 @@ import Image from './Image';
 
 import * as css from './ChallengesPanel.module.css';
 import { getReadableDate } from '../hooks';
-import { shuffleCopy } from '../utils';
+import { shuffledCopy } from '../utils';
 
 const Card = ({
   className,
@@ -71,17 +71,16 @@ const Card = ({
     </article>
   );
 };
-const MemoizedCard = memo(Card);
 
 const ChallengesPanel = ({
   challenges,
   placeholderImage,
   headerType = 'h2',
-  randomize = false
+  shuffle = false
 }) => {
   const Header = headerType;
   const [shownChallenges, _] = useState(
-    (randomize ? shuffleCopy(challenges) : challenges).slice(0, 2)
+    (shuffle ? shuffledCopy(challenges) : challenges).slice(0, 2)
   );
   return (
     <section className={css.root}>
@@ -90,15 +89,15 @@ const ChallengesPanel = ({
         <p>Suggested by the video you're watching</p>
       </div>
       <div className={css.challenges}>
-        {shownChallenges.map((challenge, key) => (
-          <Fragment key={key}>
-            <MemoizedCard
+        {shownChallenges.map((challenge, index) => (
+          <Fragment key={challenge.videoId}>
+            <Card
               className={css.challenge}
               challenge={challenge}
               placeholderImage={placeholderImage}
               headerType={`h${parseFloat(headerType[1]) + 1}`}
             />
-            {key !== challenges.length - 1 && (
+            {index !== shownChallenges.length - 1 && (
               <div className={css.spacer}></div>
             )}
           </Fragment>
@@ -108,7 +107,4 @@ const ChallengesPanel = ({
   );
 };
 
-// The ChallengesPanel component itself cannot be memoized because otherwise,
-// the shuffled array will always appear the same on every page refresh when
-// building with SSR.
-export default ChallengesPanel;
+export default memo(ChallengesPanel);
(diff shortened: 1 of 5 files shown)


shiffman:
Hi @lobertrand! This is so great! It's actually something I've been thinking about. Even though the passenger showcase scrolls it might be nice to have that randomized too. There are probably other places on the site I could imagine featuring something at random (or maybe random from a list of options, like the "featured challenges" or "featured showcase" on the landing page.)

I think the question here is whether it should be randomized when the site builds (so everything is kept completely static) or in realtime when the page is requested. I would prefer the latter but I'm not sure if the former is more appropriate for a gatsby site.

@runemadsen do you think you might be able to point us in the right direction here?

lobertrand:
I'm facing a weird behavior when suffling the array of challenges : the array of challenges seems to be shuffled on each page refresh because the title, description and dates are changing, but some parts stay static (thumbnails and links).

Looking further, it seems like **text nodes** are shuffled properly, but not **attribute values** : 
frankenstein_dynamic_content
In particular, we can see that the date text node (jan 16, 2017) matches the A* challenge, but the value of the `datetime` attribute never changes (just like the thumbnail) and matches the maze generation challenge.

This behavior only occurs when running `npm run build`, not in dev mode.

runemadsen:
Excellent! Thanks for doing this. I'm all for choosing a random challenge on the client side if we can stay under a length of `30` challenges. Since we need to load all the challenges on the client and only show one random one, it's adding `kb` to the page, but as long as we don't go wild, I think it's perfectly fine.

@lobertrand The reason why your code is not working is that you're using `Math.random()`. So when Gatsby renders the React components on the server side, it generates on random number, and when it hydrates on the client, it generates another random number.  This leads to hydration errors because React don't have matching states for the server and the client.

This is a general problem with SSR and there are multiple ways to solve it:

1. **Only render the related challenges on the client**. This is done by something like this:

```
const [relatedChallenges, setRelatedChallenges] = useState(null);

useEffect(() => {
  setRelatedChallenges([something, something]);
}, [challenges]);

return <div>
  {relatedChallenges]
</div>
```

The downside to this is that Google won't be able to index the related challenges since they are not in the initial HTML.

2. **Make a deterministic random**. This is the best solution, but it's often hard to pull off since you want a deterministic random that returns the same random numbers on the server and on the client, even when the site builds on one day on the server and is rendered on the client another day. 

Perhaps I would go with 1. for now, but you could also try 2. as a coding challenge 😂

lobertrand:
@runemadsen Thank you for your insights! I'm gonna try to get to an optimal solution :) I found this article but haven't tried it yet : https://blog.logrocket.com/fixing-gatsbys-rehydration-issue/

My work on this project is my first concrete experience with React, Gatsby, SSG/SSR and GraphQL, so I'm learning a lot from this 😁

shiffman:
Thank you @runemadsen for your super helpful and detailed reply!

@lobertrand if you need any additional support or feel you would like to hand this off to someone else, feel free to let me know anytime!

runemadsen:
Great! Let me know how it goes. I'm happy to give feedback if needed :)

lobertrand:
@shiffman @runemadsen Hi :) Here is my progress on this topic :

### 1. Rendering related challenges only on the client

I tested your first solution and it works :) The downsides of it are - like you said - that it's not ideal for SEO and that users could see this (if they scrolled fast enough or with a slow connection) :

Section appears blank and then challenges appear

### 2. Rendering related challenges both on the server and on the client

One alternative would be to render the list in some random order on SSR and in some other random order on hydration. This should be better for SEO, but users could see this :

A first shuffle appears and then is replaced with another one on each page refresh

For this to work without the problem I mentioned before, I added a ["key" attribute to force a fresh rerender](https://medium.com/@albertogasparin/forcing-state-reset-on-a-react-component-by-using-the-key-prop-14b36cd7448e) of this portion of the tree during hydration rather than letting React badly reconcile static and dynamic content together.

```jsx
const isSSR = useIsSSR();
...
<div className={css.challenges} key={isSSR}>
```

### 3. Deterministic random generator

I guess the deterministic random number generator you proposed would have to be seeded by a **different seed on each build** (and this same seed would have to be used during hydration). Otherwise, we would get the exact same shuffle on each build.

If this is feasable, this would have the advantage to not cause any flickering/loading. The only downside would be that the order would only change when a new build occurs.

By the way :
- I added the [`DEV_SSR`](https://github.com/gatsbyjs/gatsby/discussions/28138) flag to my local `gatsby-config.js` file to enable SSR in dev mode. This really helps to test this out :)
- For these screen captures, I activated the "Fast 3G" mode of Chrome DevTools to simulate a slow connection.

shiffman:
Wow! This is so cool, thank you @lobertrand! I would defer to @runemadsen in terms of which solution is optimal and best. Two questions / thoughts:

1. Is there one that feels right to you in terms of maintenance of the code?
2. Is there one that would be good in terms of setting a precedent for other similar "randomizers" on the website. For example, a "featured showcase" randomizer on the homepage or "showcase order" randomizer on challenge pages?

I'm not too concerned about SEO (tell me if I should be!) given that these are "related challenges" and not the core content of the page.

lobertrand:
@shiffman Glad to help :)

> 1. Is there one that feels right to you in terms of maintenance of the code?

The first solution seems to me like the most straightforward way to implement. The second one requires a little trick (`isSSR` boolean) but is still understandable with some comments. I think the third one would really require some hacks and complexify the understanding of the code.

> 2. Is there one that would be good in terms of setting a precedent for other similar "randomizers" on the website. For example, a "featured showcase" randomizer on the homepage or "showcase order" randomizer on challenge pages?

I guess the first one would be a great fit for secondary content that's "below the fold" (we only see it when we scroll the page so it has time to load). So it would fit the "Passenger showcase highlight" on the homepage as well as the "Passenger showcase" on the challenge page.

> I'm not too concerned about SEO (tell me if I should be!) given that these are "related challenges" and not the core content of the page.

That's what I was thinking too :)

runemadsen:
Amazing! this is great progress. I would also recommend going with the first and simplest implementation. The solution to the flicker is to render an empty div with the same height as the challenges panel when the challenges are `null`. Since the challenges panel is a set height, this will be fairly simple to do. You can even do this in cases where you don't know the height of the element you're randomizing. You'll see this on [our website](https://designsystems.international/) where we have placeholder elements before the projects are selected. There's a bit of a flicker, but it's much less than going from nothing to filled in content.

shiffman:
Wonderful! @lobertrand let me know if you need any additional support! I am so grateful for this contribution!

lobertrand:
@shiffman @runemadsen I used "empty" challenges as placeholders during the load of client side code, which gives this effect (on a slow connection) : 

skeleton_load_two

and if there is just one challenge in the "relatedChallenges" array : 

skeleton_load_one

Let me know if there is anything I can enhance :smiley: 

By the way, this component is used on two pages : the challenge page and the track video page (if related challenges exist for the given video). I applied the shuffle on both pages for simplicity, it that okay?

shiffman:
Wonderful, is this ready to review? I really like how it works. The idea is to have related challenges in all of the videos but b/c this wasn't data on the original site it wasn't prioritized during the porting (#262). Maybe I can bring this up on [tomorrow's live stream](https://www.youtube.com/watch?v=EiYhMpZ_43M).

We could even consider picking a random challenge to suggest if none are specified but maybe that's not in the spirit of this feature.

shiffman:
Ah I see your "ready to review"! I've tested it and it's good to go. @runemadsen any additional comments or thoughts before I merge?

src/utils/index.js
```diff
@@ -49,3 +50,33 @@ export const filteredPath = (resource, language, topic) => {
...
+ *
+ * @return booleans indicating whether we're in Server Side Rendering or
+ * Client Side Rendering
+ * @see {@link https://blog.logrocket.com/fixing-gatsbys-rehydration-issue/}
+ */
+export const useSSR = () => {
+  const [isServer, setIsServer] = useState(true);
+  useEffect(() => setIsServer(false), []);
+  return { isServer, isClient: !isServer };
+};
```
runemadsen, line 82: I think it would make more sense to call this hook something like `isFirstRender`, since it will render true for the first render on the server and for the first render on the client too. So something like the `isClient` boolean will confuse people because it will be `false` on the first render in the client. I propose renaming the hook to `isFirstRender` and then just returning a single boolean.

src/components/ChallengesPanel.js
```diff
@@ -77,22 +80,31 @@ const ChallengesPanel = ({
   headerType = 'h2'
 }) => {
   const Header = headerType;
+  const { isServer } = useSSR();
+  const [suggestions, setSuggestions] = useState(() =>
+    // Empty placeholders on server side render
+    challenges.map(() => ({})).slice(0, 2)
```
runemadsen, line 86: I originally thought this was a little weird, but realized that you do this to make sure there's the same number of empty challenges as there are challenges. So I would make that clear in the comment, so someone doesn't replace this with a simple `[{}, {}]`

runemadsen (approved):
I added a few small comments, but other than that it's beautiful and ready to go!

src/utils/index.js
lobertrand: You're absolutely right! I myself got confused with how many times this value was true ^^ It's much clearer now. I renamed it to `useIsFirstRender`, starting with `use` to make it clear that this is a custom hook. Since this function encapsulates a hook, [rules of hooks](https://reactjs.org/docs/hooks-rules.html) apply to it (see [this page](https://reactjs.org/docs/hooks-custom.html) for more info).

src/components/ChallengesPanel.js
lobertrand: Thanks a lot for the feedback, I added a more precise comment :)

lobertrand:
Once this is merged and if it looks good, I'd be happy to work on the implementation of the other "randomizers" we talked about :smiley:

shiffman:
Ok! Here we go @lobertrand! And yes, I would be thrilled to add more randomizers throughout the site! Feel free to DM me or open an issue to discuss further!

Issue #574, opened 2022-09-18
Licenses on projects shared via p5js sketches


alin256:
@shiffman, I did not find the license for the projects linked via sketches rather than on GitHub.
E.g., https://thecodingtrain.com/challenges/124-flocking-simulation

Would it be a good idea to include it as part of the Coding Train website?

shiffman:
Yes! Would you mind filing an issue here for discussion? Or let me know if you prefer I do! Maybe we can put something in the footer. 

https://github.com/CodingTrain/thecodingtrain.com

alin256:
> Would you mind filing an issue here for discussion?

Isn't it here now?

shiffman:
LOL oooooops, for whatever reason I thought I was in the HDD repo, I was confused! Let me remove this thread and reply correctly!

shiffman:
Yes, great idea! Maybe the footer would be a good place? Or perhaps it makes more sense to add a note to the README of the REPO given that the contents of the website itself would be under a different license. (Creative Commons Non Commercial probably).

alin256:
I tried something, but I do not have a compiler setup. Take the pull request as a suggestion only. It might be arbitrary dysfunctional, and ugly.

shiffman:
Thank you @alin256! It's my intention to use the same license for all of the code examples across the board so I don't think I need to add an additional data field to the build system. I think I can just put a note in the footer or in the GitHub README. I appreciate you thinking of this and helping propose ideas! @allcontributors add @alin256 for ideas and docs.

shiffman:
@allcontributors add @alin256 for ideas and doc

alin256:
> @allcontributors add @alin256 for ideas and doc

That is so cool :)

shiffman:
Ah, I had an idea. I think I should add info to the README of the repo and also explain the licensing in the FAQ. @runemadsen is there any license requirements I should be aware of in terms of the website codebase itself? It's my intention that the Coding Train design assets (@jasontheillustrator's illustrations etc) are creative commons non-commercial but I don't know that I've ever explicitly specified that.

runemadsen:
I think Gatsby is `The MIT License (MIT)` and so is all of the package dependencies. That's the only thing to be aware of as far as I know.

shiffman:
I was just about to close this issue, but maybe I'll leave it open as it might be worth adding a note to the website FAQ?

https://thecodingtrain.com/faq

[...]

PR #575, opened 2022-09-18
A very rough proposal for license information towards #574 

alin256:
My not-at-all-tested proposal towards #574


shiffman:
We now have #956. This doesn't explicitly cover code that is linked to from the repo (like the p5 sketches) so I suppose we could clarify that a bit better. But I think this is a good enough solution for now too.

PR #582, opened 2022-09-20
Adding 113-4d-hypercube

kfahn22:
This is ready for review. 

FYI,  I have added a link to 112 which wasn't present in original json.  Maybe unnecessary since they are sequential.


content/videos/challenges/113-4d-hypercube-aka-tesseract/index.json
```diff
@@ -0,0 +1,101 @@
...
+    "languages": ["Processing"],
+    "topics": [
+        "3d rendering",
+        "orthographic projection",
+        "rotation matrix",
+        "4d hypercube",
+        "tesseract"
+    ],
+    "canContribute": true,
+    "relatedChallenges": [],
```
shiffman, line 16: @kfahn22 we should add the 2D projection challenge here! (See #262). We could also consider doing a pass over all videos once all the challenges are ported.

kfahn22:
Sure--is the proper syntax "112-3d-rendering-with-rotation-and-projection"?  And yes, the related challenges will need to be updated after the challenges are all added.

shiffman:
> Sure--is the proper syntax "112-3d-rendering-with-rotation-and-projection"? And yes, the related challenges will need to be updated after the challenges are all added.

Correct! It's an array of challenge slugs (the last part of the URL)

shiffman:
Thank you!

PR #587, opened 2022-09-24
Fix small bug on the spacer of related challenges

lobertrand:
Hey! I just noticed that I introduced a small bug when doing the final touches on PR #572. The spacer shouldn't appear on the right :sweat_smile: 

image

So here's the fix :smile: 

image


shiffman:
Great catch!

Issue #569, opened 2022-09-14
Field to opt-in to showcase being shared on instagram (or twitter)


shiffman:
and a field for social handles? cc @kchan2000

shiffman:
Also, we should add a date! @klinegareth is working on a showcase page (see #501) and we are discussing how it would be nice to sort from most recent to oldest.

PR #588, opened 2022-09-24
Add submittedOn field to showcase contributions

dipamsen:
Added field `submittedOn` to showcase submissions, as proposed in https://github.com/CodingTrain/thecodingtrain.com/issues/569#issuecomment-1256380913.

This field contains only the date (not time) of execution of the `submission-background` netlify function.

Let me know if any changes are required!


shiffman:
Thanks for adding this! @klinegareth I think this is a great start for the showcase sorting!

PR #589, opened 2022-09-24
Cleanup to showcase files

dipamsen:
- Remove empty fields (author.url, source)
- Move source -> url if url is empty
- Replace unicode encoded escaped characters with actual characters
- Remove extra " by " from author.name
- Remove source field for p5.editor urls (redundant)


shiffman:
This is fantastic, thank you so so much!!

PR #604, opened 2022-09-29
Showcase contributions for Mandelbulb!!!

kfahn22:
I decided to port 122 quick draw but skaser's repo doesn't have a json file for that one so I went to the old site to look for the md file (after searching in vain for Rishi's script to do it automatically--where is it?).  By chance I noticed that jhrtn had contributed to the md file for mandelbulb.  I decided to have a look at it and found that the contributions that I thought were never added to the old site were in the md file (including mine).   I didn't bother with mine, but  I have added the others.


shiffman:
Thanks for catching all of these @kfahn22!!!

Issue #590, opened 2022-09-24
More randomized content


lobertrand:
Choo choo! We recently introduced randomized content on the website in the "Try a challenge!" section (on the challenge and track video page) (PR #572). Here are some other places where randomized content could make sense :

## Featured challenges on the Homepage

Currently, there can be **any number** of featured challenges and they are laid out in rows of 3 :

![featured_challenges](https://user-images.githubusercontent.com/37326143/192112316-0350187d-b2d4-41d4-9f7b-c7d641d06f03.png)

Here is the corresponding configuration ([`content/pages/homepage/index.json`](content/pages/homepage/index.json)) :
```json
  "challenges": {
    "title": "Challenges",
    "description": "...",
    "featured": [
      "174-graphics-applesoft-basic",
      "171-wave-function-collapse",
      "166-image-to-ascii",
      "126-toothpicks",
      "162-self-avoiding-walk"
    ],
    "challengesCta": {...}
  },
```

We could limit the number of challenges that are shown (to 3 or 6 for example) and pick random ones from the "featured" array.

## Featured showcase on the Homepage

Currently, there is only one featured showcase one the homepage and it changes only if we change it manually and rebuild the website :

![featured_showcase](https://user-images.githubusercontent.com/37326143/192112324-c30bb0c8-da07-492f-be70-853204bf783a.png)

Here is the configuration ([`content/pages/homepage/index.json`](content/pages/homepage/index.json)) :
```json
  "passengerShowcase": {
    "title": "Passenger showcase highlight",
    "featured": "challenges/173-snake-applesoft-basic/showcase/contribution-1661456933.json",
    "cta": {...}
  },
```
If we want to be able to pick a random contribution from a list of options, we could transform the `featured` property into an array :
```json
  "passengerShowcase": {
    "title": "Passenger showcase highlight",
    "featured": [
      "challenges/173-snake-applesoft-basic/showcase/contribution-1661456933.json",
      "challenges/102-2d-water-ripple/showcase/contribution3.json",
      "code/3-conditionals/1-conditionals/showcase/contribution3.json"
    ]
    "cta": {...}
  },
```

## Scrollable passenger showcase on the Challenge and Track video page

Currently, the items of the passenger showcase on the challenge and track video page appear in alphanumerical order of their JSON filename :

![passenger_showcase](https://user-images.githubusercontent.com/37326143/192112337-c479ab28-fa00-4bd3-9a2c-5c5550b207d5.png)

[...]

PR #606, opened 2022-09-29
Random featured Passenger Showcase on the Homepage

lobertrand:
This PR implements a part of issue #590 :steam_locomotive: 
This is a work in progress :building_construction: 

Preview : https://deploy-preview-606--codingtrain.netlify.app/

Loading on a simulated slow connection : 

passenger_showcase_highlight

### DIFF
--- a/content/pages/homepage/index.json
+++ b/content/pages/homepage/index.json
@@ -49,7 +49,9 @@
   },
   "passengerShowcase": {
     "title": "Passenger showcase highlight",
-    "featured": "challenges/173-snake-applesoft-basic/showcase/contribution-1661456933.json",
+    "featured": [
+      "challenges/173-snake-applesoft-basic/showcase/contribution-1661456933.json"
+    ],
     "cta": {
       "text": "Ready to be inspired?",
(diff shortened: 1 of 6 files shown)


lobertrand:
@shiffman Hello! I think this is ready for review :steam_locomotive: Feel free to put any passenger showcase you like in the list :smiley:
I stayed with showing only one showcase for now, as a first step.
By the way, I think think PR might interact with PR #610

shiffman:
Wonderful! This is fantastic! I've tested it and so will go ahead and merge. @runemadsen or @fdoflorenzano what do you think about us adjusting the design to feature 2 or 3 showcase projects. I'm not sure about this. Maybe that would create too much imbalance on the homepage?

runemadsen:
Awesome! I think it would be fine to have more than one passenger showcase. I think you should be able to stack them easily on top of each other, and the only problem will be the duplicated heading. You could consider rewriting it so the heading gets its own row and then you have 2-3 passenger showcase submissions listed below in rows. This way, you could keep most of the current layout.

PR #610, opened 2022-10-01
Update featured showcase

kfahn22:
There are an almost overwhelming number of really good showcases to choose from!  I picked one from 10-Print since it appears to have been one of your most popular challenges (based on number of contributions).


shiffman:
@kfahn22 would you mind redoing this pull request with the new system of having an array of showcase projects from #606?

PR #624, opened 2022-10-05
Updating 16-L system

kfahn22:
I have edited the Related Challenges, added timecodes, and added showcase images.


shiffman:
Amazing!

PR #638, opened 2022-10-07
Rough draft of showcase page

klinegareth:


fturmel:
@klinegareth @shiffman Alright - got the showcase page in a working state with a few improvements, we're back on track! It's nice to see the community contributions all in one place. You can keep iterating on the page. Let me know if you need anything or have any questions.

https://deploy-preview-638--codingtrain.netlify.app/showcase

We now have to build 12,054 pages compared to 6,686 in the last production build, and these new pages are more expensive to create. It's still manageable, but I think we'll eventually have to explore some new ideas (SSR, client-side pagination and filtering) to make the build and dev server startup times more reasonable.

fturmel:
I will also add that the language filter as currently implemented doesn't make too much sense. It refers to technologies used in the Coding Train video that inspired the contribution, and not the contribution itself.

shiffman:
Yes, I think we can do away with the filters. If I were to have filters what would be useful is:

* filter by associated video title
* filter by submitter name

I love the idea of being to see all showcase projects by any given passenger.

I could also imagine wanting to sort in reverse chronological order but that is very low priority.

node-scripts/schema.js
```diff
@@ -47,6 +48,7 @@ type Video implements VideoInterface & Node {
 
   # as resolver
   showcase: [Contribution]
+  track: Track
```
fdoflorenzano, line 51: It took me a minute to be sure, but I'm now pretty sure this could bring some inconsistencies in the data model of the site. 

From what I gather from the rest of the PR, this is needed to be able to get the URL of track videos, so each contribution links back to a video page (which I think it's nice).

The `track` reference added in the schema here kinda makes the assumption that every video is only linked to one track. But, as the data model was conceived, every track is able to reference whatever video is added into the site. Each video is their own entity currently, and every track is able to show them (including their contributions and other stuff) inside of the context of the track.

I'm not sure if currently there's already videos included in multiple tracks, but I did a test locally, adding a video in three different tracks, and the `track` reference does resolve, but just to one of the tracks. I'm guessing Gatsby makes they're best try here to resolve to something. 

So it doesn't break technically, but it may generate unexpected results. For one, the link added from the showcase from a contribution to that specific video, will only link to one of the tracks, which may be fine. But which track is it?

But we also now have the situation where a track, composed of videos, which themselves link to a track, where these two tracks may not be the same. If someone uses this reference in the future for something else, it may be get weird results.

So yeah, this may not be that bad, and we can keep an eye on this detail for future stuff.

I think there's a way of doing this "correctly", by linking each video to all tracks that includes it. And list them all in the contribution card.

fdoflorenzano (commented):
Found two things that I think are worth watching out for.

src/pages/index.js
```diff
@@ -394,13 +394,6 @@ const IndexPage = ({ data }) => {
 
         <Spacer pattern size="x2" />
 
-        <PassengerShowcaseSection
-          passengerShowcase={content.passengerShowcase}
-          placeholderImage={isFirstRender ? null : challengesPlaceholder}
-        />
-
-        <Spacer pattern size="x2" />
-
```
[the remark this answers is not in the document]
fturmel: Agreed. I can't comment on the intent for the removal though, it was already there when I started helping on this PR.

node-scripts/schema.js
fturmel: Yes, you are right, it's a valid point that it could cause some issues and confusion eventually. I was aware of this when implementing the feature, but based on the content so far, no video is ever referenced by more than one track (even though as you point out it's technically possible to do so).

I can think of two options here, either A) we enforce that a video cannot be referenced by more than one track with some additional tests in `content-testing`, or B) we update the contribution resolver to return an array of tracks.

One slight issue with option B is that starting from a contribution, we don't have enough information to locate the correct parent track for the video. We'd have to pick one at random, or link back to all matching tracks as you suggest which feels noisy and a bit odd to me.

shiffman: Just to offer some quick context here. There are videos that appear in multiple tracks! You can find challenges, for example, in several of the "side track" collections, for example: https://thecodingtrain.com/tracks/games.

However, for me, there is always a "canonical" track that the video was originally made for. For example, it might be a video I made for "Nature of Code" but also pops up in a side track about "polar coordinates." The most common scenario is a "coding challenge" that appears elsewhere in another track. The "spring forces" challenge, for example, also shows up in The Nature of Code.

I would 100% be ok with having every video paired with its "canonical" track for anything we need for the showcase page.

Does this make sense and does it help?

fturmel: @shiffman Thoughts? Is this valuable down the road that multiple tracks could potentially link to the same video? Maybe you had in mind that you could build tracks that are more like playlists?

@fdoflorenzano There is one other eventual problem here when we determine the slug from a contribution. There is an assumption that challenge videos are never linked from a track, but it's another edge case that is technically possible even though there are no such relations in the content at the moment. Nothing would break, but it would prioritize linking to a track page over a challenge page.

https://github.com/CodingTrain/thecodingtrain.com/blob/410a42c32ef7de80ca18fa9669494a52e5d46644/src/components/showcase/Card.js#L26-L28

shiffman: Upon further reflection, I see that maybe we never actually put anything in more than one place! I consider the Coding Challenges to be a "track" but I see now that it doesn't actually live as one on the site. So any challenges in other side tracks are actually just in one track.

And the spring and pendulum  challenges should be in the Nature of Code track! But we never put them there. They should be right after 3.7.

Screen Shot 2023-02-17 at 10 14 33 AM

I would like to have the option of a video being in multiple tracks (like a playlist) that was a core concept when we built the site! But I am also ok, as I previously mentioned, with there being a "canonical" track that each video is associated with, if that makes sense.

fturmel: Super helpful. So maybe we do introduce the idea of a canonical track for videos, and make it only required when we detect in `content-testing` that a video is referenced by several tracks?

And we could also add a boolean flag to the video schema to identify "challenge" videos. That would allow us to prioritize linking to the challenge page even if a challenge video is references by a track?

fdoflorenzano: That sounds great to me @fturmel! I think it's a nice way of tackling all the mentioned challenges. Maybe instead of a boolean flag, it can be a `type` string field that further defines the type of video, with `"video"` and `"challenges"` as possible values for now. Maybe someday there's another specific type of video introduced, and this is better equiped to handle that (like the Guest Tutorials we also included in the modeling but haven't really used yet).

fturmel: @shiffman Am I correct to assume that a challenge video would never need a canonical track since its challenge page would be canon no matter how many times it appears in tracks?

shiffman: Yes, I think that makes sense!

fturmel: @shiffman @fdoflorenzano Just added the canonical track feature with pretty thorough unit tests, let me know what you think!

Few things:
- There was already a `source` property on video files that I could use to identify track videos from challenge videos after all
- The tests caught a few videos referenced in more than one track, could you verify that I set the correct canonical track for these?

src/pages/index.js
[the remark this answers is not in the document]
fturmel: @shiffman @klinegareth Can you confirm if the showcase section on the homepage was supposed to be removed or was it a mistake?

content/videos/more-p5/github-pages-hosting/index.json
```diff
@@ -3,6 +3,7 @@
   "description": "How to host your p5.js sketch with GitHub pages, updated for 2020!",
   "videoNumber": "Video number",
   "videoId": "ZneWjyn18e8",
+  "canonicalTrack": "git-and-github-for-poets",
```
shiffman, line 6: I think I would consider this one to be "P5 TIPS & TRICKS" actually!

src/pages/index.js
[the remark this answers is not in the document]
shiffman: This is a mistake! The passenger showcase highlights should stay on the homepage!

node-scripts/schema.js
shiffman: I made one inline comment suggesting a different canonical track! But I'm not too worried about this as we can always change them later even if you or I mischaracterize one!

content/videos/more-p5/github-pages-hosting/index.json
fturmel: Done, I'll mark this as resolved

src/pages/index.js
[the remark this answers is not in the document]
fturmel: Should be restored now.

fturmel:
OK... that was a decent amount of work, but I extracted the filters from the large `ItemsPage` component to simplify things and allow us some flexibility. Had to also change the function signature of `filteredPath` to make it more flexible.

I implemented only filtering by authors in the showcase page after all. Mostly because 845 authors * 401 topics ~= 340k static pages to build 😂 Not going to happen unless we switch to SSR or client-side pagination/filtering.

@klinegareth I added the missing "purple" variant for the showcase filter component, but that's the only style-related change that happened over the last commits. You can pick things up from here and work on the card layout if you want. Let me know if you're stuck or need any help!

One thing to look out for are rows of 2 contributions in the 3 columns layout:
Screenshot 2023-02-18 at 9 14 33 PM

content/pages/homepage/index.json
```diff
@@ -60,9 +60,14 @@
       "challenges/133-time-tables-cardioid-visualization/showcase/contribution8.json",
       "challenges/133-time-tables-cardioid-visualization/showcase/contribution11.json"
     ],
-    "cta": {
+    "featuredCta": {
       "text": "Ready to be inspired?",
       "buttonText": "Enjoy this showcase highlight"
+    },
+    "showcaseCta": {
```
fturmel, line 67: @klinegareth I think only one of these CTAs is needed on the homepage at the moment, unless you were working on a revised passenger showcase section.

shiffman:
Wow, amazing! Thank you @fturmel! I love the filtering by author so much. I think that's really the only filter we need since to see the showcase projects for any given video I can just click over to the video page.

We are so close to merging yay!

node-scripts/schema.js
fdoflorenzano: Looks great to me! Thanks so much for this! I like this approach much better and the approach is very thorough!

shiffman:
With @klinegareth's latest commit I think this is ready to merge!!! Wowza!!! 

https://deploy-preview-638--codingtrain.netlify.app/showcase

I'm updating this with main and marking ready for review! I may do a couple final language tweaks. Any last comments or thoughts before I merge from anyone following?

I think we can make some refinements going forward:

* handle video titles that are cut off
* add information / headers to each passenger filter page so it's a bit more like a "profile" page
* change author links to go to their filtered page (and there, you'd find a link to their website, etc.)?

We can open a new issue to track!

fturmel:
@shiffman Looks good to me, and I agree with the refinement ideas.

I would add that we might want a "browse showcase" button on the homepage, just like tracks and challenges.
image

fturmel:
Oh and we haven't resolved the homepage JSON with two CTA comment I left before. Would be nice to fix before we merge and just forget.

shiffman:
@fturmel I think I cleaned this up in 07a2b8d3771e613d39643174962280115812efa3, how does it look?

fturmel:
@shiffman Perfect!

shiffman:
Here we gooooooooooooooo!

PR #642, opened 2022-10-08
Adding related challenges 

kfahn22:
This pull request adds related challenges, updates some code links that went to the archived repo, and adds timestamps for 90-dithering.


shiffman:
This is so fantastic! WOW!

I'm realizing we should probably do a similar pass with the tracks too! So much content and data!

Thank you hero @kfahn22!

PR #646, opened 2022-10-09
A Showcase for coding challange 124: flocking fire helicopters

alin256:
Trying to properly participate in Hacktober Fest while submitting my showcase
The showcase shows:

* [A flock of LEGO fire helicopters that my son dreams of assembling](https://www.youtube.com/watch?v=roD93qUTQ7o)
* The flocking physics has been seriously changed, 
* But the interesting thing is the fire helicopter animation.


shiffman:
Noting that we should also assign the label `showcase` if the pull requests comes via GitHub directly! I use the label to sort and look at recent contributions!

PR #640, opened 2022-10-08
Adding 111-animated-sprites

borntofrappe:
Following the example of #103 I've tried to summarize the changes in a fewer commits.

- [x] updated contributions to replace missing pages and add images
- [x] updated index.json with the topics, links, and image for the challenge, icons and descriptions for links

PR #648, opened 2022-10-10
Adding 66-javascript-countdown-timer

borntofrappe:


borntofrappe:
For `javascript-countdown-timer.jpg` I included a picture of the timer alongside the JavaScript code in the p5 web editor, I hope it's satisfactory. 

Ultimately I created a similar version with a more appropriate aspect ratio for `index.jpg`. Let me know if the image could work for `index.jpg` as well (similarly to #640).

javascript-countdown-timer

Unfortunately the output of the challenge is not visually striking, but there's a lot of knowledge behind that string of text.

borntofrappe:
It's a pleasure @kfahn22 :)

I will undoubtedly use the thumbnail you proposed, but I have one last question: should the image have the same constraints as those for the contributions? 800pixels wide, at most? The weiught is already well below the 500+kb threshold.

kfahn22:
No, the index images do not have the same constraints and ideally should 1280 by 720.

shiffman:
Amazing!!!

PR #651, opened 2022-10-11
Images and related challenges for 78 and 103

kfahn22:
I am not sure adding color for 103 was a good idea, so let me know if you would prefer black and white.


shiffman:
I like the color for 103, but I wonder if something more zoomed in to just the fire effect would make more sense. Not my finest work but something along the lines of. . .

image

This one is tricky!

kfahn22:
What is the resolution of your image?  I have added a new one, but it is really low res.

shiffman:
> What is the resolution of your image? I have added a new one, but it is really low res.

I scaled up the canvas just messing around, let's try yours and see how it looks!

PR #659, opened 2022-10-13
yt-desc updates and cleanup

dipamsen:
- Added `relatedChallenges` to `yt-desc`
- Cleaning up invalid links and references
- Update `yt-desc` to prefer yt urls over website urls (Hope it is not too late for this...)
- Add `playlistId`s to tracks


shiffman:
This is awesome yay!!! I'll merge once the checks complete and I'll click around the preview just in case.

PR #664, opened 2022-10-13
Adding 60-butterfly-generator

borntofrappe:
Adding challenge 60.

Unfortunately there's no image for the contribution.

For `index.jpg` I tinkered with the p5.js web editor to create a 1280x720 image with known colors, but I have a question. Since the asset is picked up for the code example as well (there's no reference to a different asset in the "codeExamples" field), should the image be at most 800 pixel wide, like the assets for the contributions?

I created an alternative just in case.

Butterfly Generator thumbnail

Let me know.


DenisovichDev:
Hey everything looks great! I don't think you need to add the same image in a different dimension if you wish to use the same image. Only this one will do. Great work!

shiffman:
I'll go ahead and merge! We can remove the extra image separately if we want! THank you @borntofrappe!!

PR #665, opened 2022-10-13
Multiple Featured Showcase on the Homepage

lobertrand:
Hi! :steam_locomotive:  This PR implements a part of issue #590. It's a rework on the Passenger Showcase Highlight section on the Homepage. The goal is to display three contributions instead of one.

The simplest way to stack 3 contributions woud be to do this :
version1

I worked a bit more on this design so that it fits better with the rest of the page : 
1. I used a bigger font-size for the "PASSENGER SHOWCASE HIGHLIGHT" title do that it fits better with the rest of the titles on the Homepage ;
2. I replaced the three "READY TO BE INSPIRED?" banners with only one at the top to avoid duplication and to avoid having three big CTA buttons at the same place ;
3. I added hierarchy on the details of the contribution by playing with the font sizes and colors : I emphasized the title of the contribution, then the author and finally the source video ;
4. I let the image take 40% percent of the width instead of 25%, so fill the empty space an give more presence and a "cinematic" feel to the image ;
5. To replace the CTA, I added a play button in the corner of the images, just like on the challenge and track video pages, this also adds consistency.

version2
Preview : https://deploy-preview-665--codingtrain.netlify.app/

These are only ideas and I'm open to any rework. Feel free to give me your opinion on these changes :smiley: @shiffman @runemadsen

### DIFF
--- a/content/pages/homepage/index.json
+++ b/content/pages/homepage/index.json
@@ -48,7 +48,7 @@
     }
   },
   "passengerShowcase": {
-    "title": "Passenger showcase highlight",
+    "title": "Passenger showcase highlights",
     "featured": [
       "challenges/76-10Print/showcase/contribution41.json",
       "challenges/126-toothpicks/showcase/contribution1.json",
--- a/src/pages/index.js
+++ b/src/pages/index.js
@@ -107,16 +107,26 @@ const PassengerShowcaseCard = ({ showcase, placeholderImage, cta }) => {
           </p>
           <p className={css.showcaseTitle}>{title}</p>
         </div>
-        <address className={css.author}>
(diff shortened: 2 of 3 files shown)


shiffman:
Thank you for this @lobertrand! It's very exciting to see! I'll let @runemadsen weigh in in case he has any thoughts about the design, I like the adjustments you've made!

shiffman:
Oh and let's make `HIGHLIGHT` plural at the top!

runemadsen:
I think this is REALLY lovely! Code is perfect too. Thanks so much for the clear PR and for working within the system. I vote merge!

shiffman:
Allright, here we go!! Thank you @lobertrand, incredible work!

PR #673, opened 2022-10-15
Removes AEOUD From Upcoming Events

DenisovichDev:
Removes An Evening Of Unnecessary Detail from the upcoming events list. Ignore the commits, it's a stupid git mistake.


shiffman:
Thank you! Though I'm sad to take this away, I had so much fun!

PR #675, opened 2022-10-15
Porting CC 47 Pixel Sorting in Processing

anusha-exe:
Things that have been done:

- checked and cleaned up links and tags
- added timestamps
- updated the thumbnail according to the requirements. I don't know if this is okay. Feel free to change it as you see fit.
- added screenshot of the showcase
Related challenges section needs to be updated since Quick Sort Visualization hasn't been ported yet.


kfahn22:
Thank you!!  I have a couple of updates, otherwise this looks great!  

I would use the image of sunflowers for the code image.  I can update that if you would prefer.

The link to the Processing code should be updated  because the prior repo is now archived.  Here is the new link.
https://github.com/CodingTrain/Coding-Challenges/tree/main/047_PixelSorting/Processing

anusha-exe:
@kfahn22  Thank you for the corrections! I have changed the index.jpg and the Processing code link. Please check if everything's fine now

kfahn22:
Thank you, this looks great!  I apologize, I loved your original index image!   I was only suggested that you  add a code example image since the png file in the code sketch is sunflowers but I didn't make that very clear.

anusha-exe:
Ah, I didn't realize. I'll make the changes, no problem

shiffman:
Thank you for this wonderful contribution!!

I love that you are creatively trying a new image for this and using the Coding Train background! I think that image, however, doesn't demonstrate pixel sorting so well because the source image has something of a sorted gradient sky. So I might suggest using he sunflower or else do you want to try a different Coding Train specific image?

kfahn22:
pixel
Sorry for the extra work I have caused you!  Here is an image of Daniel and Gloria that you can  use if you want.

anusha-exe:
@kfahn22 I don't mind that at all! How can anyone keep anything but this after looking at this image TT. Thank you! I'll be using this then, @shiffman

PR #678, opened 2022-10-16
Adds back the lost Clappy Bird images

DenisovichDev:
This PR adds the deleted images from #663


shiffman:
Thank you! Apologies for messing this up in the first place!

PR #685, opened 2022-10-18
Adding 48 white house social media data visualization

borntofrappe:
Adding this challenge requires an update to the p5.js sketch. I will try to summarize my findings in a following comment.


borntofrappe:
The starter template [for the coding challenge](https://github.com/skaser85/thecodingtrain_website_data/tree/main/challenges/48-white-house-social-media-data-visualization) has only a single file for the image thumbnail. Looking at [the archived website](https://codingtrain.github.io/website-archive/CodingChallenges/048-tweetsbymonth.html) I was able to find a reference to [the p5.js sketch](https://editor.p5js.org/codingtrain/sketches/adc42eDiu), but running the sketch produces a rather unusual result.

p5 Sketch Issue

Looking at the code  in `sketch.js` the issue is with line 39, which creates a date object from the string stored in the `timestamp` property.

```js
var date = new Date(tweets[i].timestamp);
```

For reference a string has the following format `2016-11-14 18:33:08 +0000`. Passing the string as-is produces an invalid date.

```js
new Date('2016-11-14 18:33:08 +0000'); // Invalid Date
```

Since the date is then used to create a key for the tweets, the code fails to create a unique key. All tweets share the same identifier and you end up with a collection of 1. One way to fix this is to break down the string into separate arguments, as per the syntax highlighted [in the MDN docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/Date#syntax
)

```js
// new Date(year, monthIndex, day, hours, minutes, seconds,)
new Date(2016,11,14,18,33,08); // Date Wed Dec 14 2016 18:33:08
```

Ultimately this means the sketch can be fixed updating line 39 as follows.

```diff
-var date = new Date(tweets[i].timestamp);
+var date = new Date(...tweets[i].timestamp.split(/[\: -]/));
```

In the process you lose the precision of the time zone (`+0000`), but the line works to plot the data.

p5 Sketch Fixed

borntofrappe:
Apologies if I was too verbose in the explanation. The short version is: if we look at the sketch from the archived website.

https://editor.p5js.org/codingtrain/sketches/adc42eDiu

We need to update line 39 in `sketch.js` to ensure the date is a valid date object. One way to achieve this is as follows.

```diff
-var date = new Date(tweets[i].timestamp);
+var date = new Date(...tweets[i].timestamp.split(/[\: -]/));
```

Once the sketch works I will add a reference in `index.json`. Alongside the timestamps (I am working on those), that should wrap up the pull request.

borntofrappe:
**Update**: looking at the video it seems the current sketch works on Chrome, but not on Firefox 🤯 

Apologies on top of my previous apologies...it seems I've stumbled on a quirky cross-browser issue.

With Chrome

Chrome date

With Firefox

Firefox date

I'll leave it up to you then. Let me know how to proceed. Again, sorry for quite the messy development.

DenisovichDev:
Hey @borntofrappe! The code works for me, so this could be a cross browser issue. Which browser are you using?

DenisovichDev:
@borntofrappe Even if this is a cross browser issue, it is worth changing and noting in the source. I would say the best thing to do would be to update your change in the [coding challenge repo](https://github.com/CodingTrain/Coding-Challenges), since your change works in both chrome and firefox. So if it's no problem, make a PR to the challenge repository with your suggested changes, preferably with a comment explaining why it's different from the video. Thank you so much!!

I would also suggest you to change the p5 link in `codeExamples` to the web editor link, since that's what's being followed usually. Great work!

borntofrappe:
> I would also suggest you to change the p5 link in `codeExamples` to the web editor link, since that's what's being followed usually. Great work!

Thanks :)

For the link I was waiting for confirmation, as perhaps you wanted to create a different sketch. If it's okay I'll include the one mentioned above though.

DenisovichDev:
> For the link I was waiting for confirmation, as perhaps you wanted to create a different sketch. If it's okay I'll include the one mentioned above though.

Awesome. If you mean this one: https://editor.p5js.org/codingtrain/sketches/adc42eDiu go ahead! Maybe @shiffman would like to change the line in the web editor sketch for firefox compatibility, but that's something that can always be done later anyway. So I don't see any harm adding the link for now.

kfahn22:
The sketch has been updated with the code you provided.  Thank you so much!

kfahn22:
obama
Here is an image with the words.  You might be able to come up with a better version, but you can use this one if you want.

kfahn22:
Sorry about this!  I have another version which might be better, but you  can decide.
obama_words_at-bottom

borntofrappe:
I've added a bit of padding around the bars, but otherwise I've kept the image intact. The font is a much better choice than the default sans-serif.

shiffman:
Thanks for all of this work and discussion! I added a comment in the web editor version to explain the addition of `...`. I really need to do a video about the spread operator! That's the only way I'll actually remember how it works myself 😂

PR #686, opened 2022-10-18
Porting CC 089 Langton's Ant

anusha-exe:
Added:

1. Thumbnail according to the CT color scheme
2.  Timestamps
3.  Showcase screenshots
4.  Descriptions, tags, icons
5.  Checked links


content/videos/challenges/89-langtons-ant/index.json
```diff
@@ -0,0 +1,90 @@
+{
+    "title": "Langton's Ant",
+    "description": "In this coding challenge, I implement Langton\u2019s Ant in Processing, which is a cellular automaton that starts off with simples rules but takes on complex emergent behavior.",
```
shiffman, line 3:
```suggestion
    "description": "Langton's Ant is a cellular automaton that starts off with simples rules but takes on complex emergent behavior.",
```

kfahn22:
This looks great!  I think that "102-2d-water-ripple" might be another challenge to add to related challenges.

PR #690, opened 2022-10-19
YT Description Generator Bug Fix

shiffman:
@dipamsen can you check this? cc @kfahn22

### DIFF
--- a/node-scripts/yt-description.js
+++ b/node-scripts/yt-description.js
@@ -321,10 +321,10 @@ function writeDescription(video) {
         (vid) => vid.pageURL === `challenges/${challenge}`
       );
       if (challengeData) {
-        const { videoNumber, title } = challengeData.data;
+        const { videoNumber, challengeTitle } = challengeData.data;
         const url = challengeData.pageURL;
         description +=
-          `🚂 #${videoNumber} ${title}: ${getYouTubeURL(url)}`.trim() + '\n';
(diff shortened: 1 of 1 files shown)


node-scripts/yt-description.js
```diff
@@ -313,18 +313,16 @@ function writeDescription(video) {
...
       if (challengeData) {
-        const {
-          videoNumber,
-          challengeTitle: title,
-          pageURL: url
-        } = challengeData.data;
+        const { videoNumber, title } = challengeData.data;
+        const url = challengeData.pageURL;
         description +=
           `🚂 #${videoNumber} ${title}: ${getYouTubeURL(url)}`.trim() + '\n';
```
dipamsen, line 327:
```suggestion
        const { videoNumber, challengeTitle } = challengeData.data;
        const url = challengeData.pageURL;
        description +=
          `🚂 #${videoNumber} ${challengeTitle}: ${getYouTubeURL(url)}`.trim() + '\n';
```

This fixes the annoying bug of showing `Asteroids - Part 1` for the title. 

This is a hacky fix for multipart challenges which I have created wherein the `title` also contains the Part information, but the `challengeTitle` is just the name of the challenge. This is required as in different places different versions are needed.

dipamsen (changes requested):
This looks great! Just suggested one change.

shiffman:
Thanks @dipamsen! I think it's fine to link to just part 1 of the challenge. I suppose we could consider linking to thecodingtrain.com page itself but maybe for YT's metadata/algorithms better to link to YT on YT itself!

PR #693, opened 2022-10-19
Porting CC 137 4D OpenSimplex Noise Loop

anusha-exe:
Added:
1.  Timestamps
2.  Descriptions, related challenges, language and topic tags, icons
3.  Replaced index.jpg with new thumbnail conforming to the CT color scheme
4.  Cleaned up group links

### A Note:
This challenge had only one contribution by [Juan Carlos](https://www.jcponce.com/p/about.html). His sketch is on flocking simulation of knot like structures using OpenSimplex noise, and since it was unrelated to gif loops or noise loops, I have removed it. His sketch:
https://editor.p5js.org/jcponce/sketches/4cmRPWfF
I think it would be preferrable to add it to the `124-flocking-simulation`? @shiffman


shiffman:
> https://editor.p5js.org/jcponce/sketches/4cmRPWfF
> I think it would be preferrable to add it to the 124-flocking-simulation? @shiffman

@anusha-exe I think this actually uses Simplex Noise to define the shapes of the flocking elements so it's good to have here!

anusha-exe:
Ohh, I see. I should've checked the sketch more thoroughly. I'll add it back, then!

shiffman:
Fantastic!

PR #697, opened 2022-10-20
Adding 109 visualizing 500 000 subscribers

borntofrappe:
Let me know if you have improvements in terms of thumbnail. I've used the familiar purple for the circles, but for the map there is less freedom being loaded through Mappa.js and Leaflet.


DenisovichDev:
Thank you so much for all the effort you have put in this!! Great work on the timestamps!
The thumbnail looks good to me, I don't see any harm using it.

kfahn22:
I agree, great job!

shiffman:
Wow, incredilble!

PR #703, opened 2022-10-21
fix typo in url

dipamsen:
(not updating related challenges as it is claimed by @kfahn22 in the spreadsheet)


kfahn22:
If you think other challenges could be added to relatedChallenges, please add!

dipamsen:
No, I meant the editing part wherein the challenges from the `Videos` section are moved to `relatedChallenges`, so here 162 can be added to the top and the videos section can be removed. @kfahn22 
I am not doing it to prevent overlap as you have been doing this now for many challnegs.

shiffman:
Thanks for catching this!!

PR #705, opened 2022-10-21
Fix height of passenger showcase details on homepage

lobertrand:
Hi :steam_locomotive: I just noticed some overflow on the title of passenger showcases on the homepage (PR #665). Here's a fix :

Wide screen (before)
wide-before
Wide screen (after)
wide-after

Medium screen (before)
medium-before
Medium screen (after)
medium-after

Small screen (before)
small-before
Small screen (after)
small-after


shiffman:
Great catch!

PR #708, opened 2022-10-21
Adding CC 063 Texturing Cloth Simulation (Unikitty Flag)

anusha-exe:
Added:

-  Timestamps of both parts
-  Added related challenges, language, topic tags
-  Added descriptions, icons, thumbnails
-  Cleaned code examples, references, added new video links (livestream archive, tutorial playlist)

Ready for [preview](https://deploy-preview-708--codingtrain.netlify.app/challenges/3)

### Notes

- There is no code example for the first part of the challenge
- The toxiclibs website has been closed down now. From the CC 020, I have updated the links and the description accordingly. 
-  The challenge is called Texturing Cloth Simulation but the code example file is saved in `063_unikitty_flag/Processing/CC_063_unikitty_flag` which might lead to confusion? But since the challenge number stays the same, I guess it's fine. Mentioning this just in case.
-  I have kept the thumbnail with unikitty (although the image is square and the title doesn't mention unikitty) because the drawing is so sweet. I'll leave it to @kfahn22 in case something better can be done about it.
-  The description for the challenge became quite big.


kfahn22:
I agree that Daniel's daughters drawing is the best!!

shiffman:
I think this one is actually my son's! 🎨

DenisovichDev:
Very sweet!

anusha-exe:
Yes, when he was an eight year old! :D

PR #709, opened 2022-10-22
42 markov chain name generator

borntofrappe:
I'm in the process of creating the timestamps, but I thought of creating a pull request first to double-check the images and index.json.

Being a multi-part challenge the structure is more complex.


DenisovichDev:
I think your thumbnail looks great! If I were to nitpick, one thing I could point out about the `markov-chain-name-generator.jpg` file is that it's better if the image is not a square and is in some sort of landscape aspect ratio, so that it shows better when it is cropped to fit. But beside, everything looks good!

Also, if you are still working on the PR, you could mark it as a "draft pull request" to let everyone know that it is a work in progress. Thanks!

kfahn22:
Your image looks great!

I wanted to add that the convention for the index images is 1280 by 720.  One additional consideration is that the edges of the images are clipped on some devices (not a problem with yours as far as I could tell), so I am started to redo some to give padding around edges.

borntofrappe:
Understood, thanks for the guidance @DenisovichDev @kfahn22

borntofrappe:
I've updated the images for the sketches and included the timestamps, which should cover everything.

Markov chains are quite intriguing. Thanks for the challenge, literally.

DenisovichDev:
Thank you so much for all your efforts!! Everything looks great!

shiffman:
Awesome!

PR #714, opened 2022-10-24
Adding 44 afinn 111 sentiment analysis

borntofrappe:
The live demo is quite heavy on text, so I tried to summarize the most important visuals in `index.jpg`.

index image

Hopefully it is enough to illustrate the technique with the AFINN list.


anusha-exe:
I'd like to mention a little thing, if you don't mind. The commas should have been after the number with the space following it. `great: 2, good: 3`

anusha-exe:
I'd like to mention a little thing, if you don't mind. The commas should have been after the number with the space following it. `great: 3, good: 3`

borntofrappe:
Thanks for the mention @anusha-exe. It makes a world of a difference

shiffman:
This looks so good thank you!!

anusha-exe:
Glad to be of help! The thumbnail looks great

PR #722, opened 2022-10-24
Update maintain-the-website guide to include info about adding links to Live Stream archives 

kfahn22:
I have added some info about adding the links to livestreams and included an image.


shiffman:
Thank you @kfahn22!

PR #713, opened 2022-10-23
Porting CC 043 Context Free Grammar Generator

anusha-exe:
🔨 Added: 

-  timestamps
-  links to code example, references, tutorial playlist, livestream archive
-  descriptions, icons, tags
-  thumbnails

Ready for [preview](https://deploy-preview-713--codingtrain.netlify.app/challenges/3) 🚂 

## 📋  Notes:

1. This challenge had no generated JSON so I have added what I found from the YouTube videos and whatever felt right to me. So, the description and the links needs to be checked and if anything is missing or unnecessary, please tell me.
2.  From what I saw, this challenge had no showcase contributions either. 
~3.  It is an old challenge so there isn't any web editor p5.js version so I have kept the link to the repo in the `p5:` field. If you can create a web editor version, then I'll update the links~
4.  I added new words to the text generator. I hope this thumbnails are okay. ~Perhaps the font size is too small and unreadable? And the original thumbnail from `images` is also not legible it seems~
5.  This CC is a part of the playlist on Context-Free-Grammar. That playlist is a prospective sidetrack. If no one has started working on it, I could do it

PR #724, opened 2022-10-25
Adding 79 number guessing chatbot

borntofrappe:
- adds challenge files
- updates contributions
- adds (and updates) thumbnail
- updates challenge details (timestamps, references, videos, livestream)


DenisovichDev:
Hey! Thank you for another submission! I'd like to mention a small thing. Judging from the convention that is being followed regarding the Live Stream Archive, since there is just one Live Stream VOD in this case it doesn't need a different `groupLinks` category, it can very well be under `videos`. And according to the discussions in #713, the title could simply be "Live Stream Archive" (since it's just one video) and you can add the live stream # in the description!

Other than that, everything looks great! I'll cc @kfahn22 for any suggestions about `relatedChallenges`

borntofrappe:
Understood @DenisovichDev 

Reviewing the video the only related challenge would be # 80, but it is more likely that the relation goes the other way around (since 80 builds on top of the current challenge)

kfahn22:
Thank you! 
Since there is a link to the Programming with Text playlist, I would suggest adding `"37-diastic-machine"` , `"39-madlibs-generator"` for related challenges for now.  There are some others that haven't been ported yet.

PR #729, opened 2022-10-26
Adding CC 080 Voice Chatbot with p5.Speech

anusha-exe:
🔨 Added: 

-  timestamps
-  links to code example, references, tutorial playlist, livestream archive
-  descriptions, icons, tags
-  thumbnail
-  showcase contribution 

Ready for [preview](https://deploy-preview-729--codingtrain.netlify.app/challenges/2) 🚂

## 📋  Notes:

-  Challenges 79 and 80 would be more suitable as parts of the same challenge, maybe?
-  Related challenges? The nearest to speech recognition would be perhaps the challenges on neural networks and machine learning.
-  There's one contribution in this challenge but the [live preview](https://davebsoft.com/software/stupid-nonsense-spoken-sentences/) didn't work properly for me and on hosting the source code on local server, the chatbot kept repeating itself and wasn't taking input from the user's speech. Hence, the screenshot I added doesn't show any input or its results.
- I added RiverScript as one of the tags for `languages`. Please tell me if I should keep it.
-  For the thumbnail I was thinking of something like this. Although it doesn't necessarily indicate it's speech and not text.
 
index


DenisovichDev:
Hey! Everything looks great! Just one thing would be to change the archived p5 link to https://github.com/CodingTrain/Coding-Challenges/tree/main/080_Voice_Chatbot_with_p5.Speech/P5 instead.
- About 79, it probably makes sense, but since those have different values for the `videoNumber` properties, it's best to keep them separate. 79 could actually be added as a `relatedChallenge` later, in that case.
- It works! You could choose to not add the thumbnail for a showcase, in that case the `index.json` would be used as a placeholder in the build.
- It's good! You can keep it.
- I love the thumbnail!! I think it's great!

anusha-exe:
Okay, thank you for the clarifications

shiffman:
Agreed on all points, thank you @anusha-exe and @DenisovichDev!!

PR #730, opened 2022-10-26
Cleanup

dipamsen:
The monthly showcase cleanup pass is here :)

- removed empty url and source fields
- removed unnecessary p5 web editor source fields
- Changed `videoNumber` for cabana videos to make consistent with other challenges


shiffman:
Wow, thank you @dipamsen!!!

PR #731, opened 2022-10-26
Adding CC 82: The Ex-kitten-sion (Image Replacement Chrome Ext)

dipamsen:
Deploy Preview: https://deploy-preview-731--codingtrain.netlify.app/challenges/82-image-chrome-extension

image

- ~I could not think of a good thumbnail image, so used the Rainbow wiki page for now. (also it is has a bad aspect ratio)~


dipamsen:
added ls archive, waiting on rc for 83 and 84 (i'll add these next)

dipamsen:
added thumbnail as suggested by @kfahn22 with some changes.

shiffman:
This is great!

PR #736, opened 2022-10-27
Add times to LS links in challenges (CC1-20)

dipamsen:
Adding timecodes (`t=ABCs`) so that the livestream link starts at the start of the recording of the challenge.

- challenges 1 to 20


dipamsen:
This can be merged, I'll be updating these in chunks to minimise conflicts.

PR #737, opened 2022-10-27
Some fixes

anusha-exe:
A little PR
- Since CC 79 and 80 are like parts of the same challenge, I made their topic and language tags same
-  Changed group link titles of CC 108 and CC 002.
I don't think it needs to have the `hacktober-accepted` label.


shiffman:
Great!

Issue #738, opened 2022-10-27
Multi-Part Challenges break a side track


shiffman:
At the moment, we cannot include multi-part challenges in a side track, see an example here:

https://thecodingtrain.com/tracks/games/46-asteroids

@lobertrand what do you think the best way to handle this is? 

@kfahn22 do you think we should remove them temporarily until we have this resolved?

lobertrand:
Hello, I guess one way to handle this could "spread" the multiple parts on the track, something like this : 
![image](https://user-images.githubusercontent.com/37326143/198372798-a563f172-ecb8-4de2-83b4-12b4f77c21bc.png)
I don't know how hard that would be to implement but I'm willing to try :smiley:

kfahn22:
Yes I will remove

shiffman:
Once this is resolved we can revert #739

shiffman:
@lobertrand this seems like a great solution to me!

lobertrand:
Hi @shiffman @kfahn22 :steam_locomotive: 
I'm working on fixing this issue and I was wondering which title I should display in the timeline for multi-part challenges.
If we take this example : 
```json
{
  "title": "Agar.io",
  "parts": [
    { "title": "Agar.io - Part 1 (Basic Game Mechanics)" },
    { "title": "Agar.io - Part 2 (Networking with Socket.IO and Node.js)" }
  ]
}
```
Which way should we display the title?
A. `video.parts[*].title`
![image](https://user-images.githubusercontent.com/37326143/198888045-bdaea5b9-7993-43d0-a1b2-fafb4364ae3e.png)
B. `video.title - Part #`
![image](https://user-images.githubusercontent.com/37326143/198888084-7b1531da-e46e-40f1-a9b5-8c63864539fa.png)

Option A works great with the above example but less with the following ones: 
```json
{
  "title": "Polar Noise Loops",
  "parts": [
    { "title": "Polar Perlin Noise Loops" },
    { "title": "Perlin Noise GIF Loops" }
  ]
}
```
```json
{
  "title": "Traveling Salesperson",
  "parts": [
    { "title": "Part 1: Traveling Salesperson (TSP)" },
    { "title": "Part 2: Lexicographic Order" },
    { "title": "Part 3: TSP with Lexicographic Order" },
    { "title": "Part 4: TSP with Genetic Algorithm" },
    { "title": "Part 5: TSP with Genetic Algorithm and Crossover" }
  ]
}
```
[...]

PR #739, opened 2022-10-27
Removing multi-part challenges from games track

kfahn22:
agario
asteroids
frogger


shiffman:
Let's plan to add these back once we resolve #738

PR #740, opened 2022-10-28
Add CC 83: Doodler Chrome Extension

dipamsen:
image

🚂[Deploy Preview](https://deploy-preview-740--codingtrain.netlify.app/challenges/83-p5-sketch-chrome-extension)

(The thumbnail image is just a screenshot from the video; someone will surely create a better creative thumbnail later)


kfahn22:
I will ask my daughter (who is an expert doodler) to come up with something.

dipamsen:
The aspect for these extension challenges which is difficult to decide is on what website the image should be created. I was trying on the CT main page, but it is already filled multiple illustrations, it didn't make much sense :/

kfahn22:
I am wondering if we can use the Nature of Code website, and then have my daughter add physics related doodles.  (Not sure exactly which screenshot would be best.)
noc

shiffman:
@kfahn22 The nature of code website is being redesigned! What about the p5 homepage?

shiffman:
I'm going to merge this and we can replace the image later!

PR #743, opened 2022-10-29
Handle Multi-Part Challenges in Tracks

### LINKED ISSUE #738: Multi-Part Challenges break a side track

lobertrand:
This fixes issue #738 :steam_locomotive: 

Preview: https://deploy-preview-743--codingtrain.netlify.app/tracks/games/46-asteroids


shiffman:
Loooks fantastic!

PR #747, opened 2022-10-30
Featured challenges/ showcase-10/30

kfahn22:
update featured challenges(again randomly chosen) and featured showcases
Updating name on one of the featured showcases


shiffman:
❤️

PR #766, opened 2022-11-07
Updating featured challenges/showcases 

kfahn22:
Updating homepage 11/7 and eliminating filler in showcase jsons.


shiffman:
Thanks for catching these!

PR #773, opened 2022-11-12
Showcase Cleanup

dipamsen:
Showcase: moving source to url, removing source property, removing empty fields

Also fixed some URLs which pointed to the archived repo.


shiffman:
Thanks for this!

PR #489, opened 2022-08-24
Begin Cleaning Up Duplicate Languages on Dropdowns

M0nica:
## Summary 
This pull request aims to resolve the duplicate languages appearing in the dropdown that were reported in: https://github.com/CodingTrain/thecodingtrain.com/issues/488

- Updated to make `languages` a set that only displays one lower cased version of each language. 

## Outstanding Questions
I'm not sure if we would prefer to have an object that we can use as a dictionary to normalize the language names while maintaining certain language-specific casing like `JavaScript` while also avoiding duplicates.  

## TODO:
- [x] handle duplicates with the same spelling but different casing (i.e, Processing vs. processing)
- [ ] handle duplicate languages with spelling variations (i.e., P5.js and P5js) in slugs and dropdown

PR #781, opened 2022-11-19
Build performance improvements

fturmel:
Hi @shiffman & team!

I was watching the [stream on YouTube](https://youtu.be/EzYDNydDUfc?t=2391) last Friday and was intrigued by the Gatsby/GraphQL/Netlify log warnings and slow builds. Thought it would be a fun challenge to learn a bit more about these and see if I could help out.

The slowness comes from building the tracks and challenges pages with all filter permutations and pagination. It ends up being a lot of pages!

Here's what I did:

- **Make some of the async code run serially**. There are just too many things running at once, and it ends up making performance worst. This was enough to get rid of the flood of warnings in the logs.
- **Rewrote some GraphQL queries** to make them a bit faster
- **Build contextual languages and topics lists to reduce total permutations and improve UX**. While the challenges and tracks share some topics and languages, there's a lot that doesn't overlap here. This means lots of useless result pages were being generated, and dropdown filters had a lot of options that would return no matches. This appears to have cut the amount of generated pages by more than half.
- **Re-architect how tracks and challenges filters are populated**. We are now generating two static JSON files that are fetched at runtime, which has a lot of positive repercussions:
  - less work for `createPages` process
  - smaller `page-data.json` for every page with filters, and the extracted data found in those new JSON files can be cached by the browser during a user session
  - if there is a topic or language content change, I think that only the relevant pages would get rebuilt instead of every single page that has a filter dropdown component

Some stats on my M1 MacBook Air:

|| total pages built | `createPages` execution |
| --- | ---: | ---: |
|baseline| 19,911 | 51 seconds |
|with this PR| 7,213 | 9 seconds |

**Bonus!**
- Added the `gatsby-plugin-netlify` plugin dependency. It seems to log slow queries during build, so potentially some good information to optimize things further.
- Fixed a few typos in the content, mostly to observe the Netlify build behavior
- Fixed a caching issue. Changes to topics and languages inside track videos and chapters did not bubble up correctly to the parent track level until the Gatsby cache expired or was explicitly purged. This was fixed by using GraphQL resolvers. It should improve the cache deployment issues mentioned in https://github.com/CodingTrain/thecodingtrain.com/pull/786#issuecomment-1322058585

Lastly, I noticed that the Gatsby Netlify plugin is out of date. Maybe that can further help with build speed and caching? I think it's called `Essential Gatsby` in the dashboard. It needs to be uninstalled and re-installed to upgrade https://app.netlify.com/plugins

Screenshot 2022-11-19 at 1 36 52 PM

Hope that all made sense and I didn't break anything in the process!

Let me know if I can help with something else. Perhaps wrapping up #489 since it would also help reduce the overall amount of built pages by removing duplicated topics and languages?

### DIFF
--- a/content/videos/challenges/153-interactive-drawing-with-sketchrnn/index.json
+++ b/content/videos/challenges/153-interactive-drawing-with-sketchrnn/index.json
@@ -4,7 +4,7 @@
   "videoNumber": "153",
   "videoId": "ZCXkvwLxBrA",
   "date": "2019-10-31",
-  "languages": ["ml5.js", " p5.js"],
+  "languages": ["ml5.js", "p5.js"],
   "topics": ["sketchRNN", "machine Learning", "Quick, Draw!"],
   "canContribute": true,
   "relatedChallenges": ["128-sketch-rnn-snowflakes", "152-rdp-algorithm"],
@@ -83,12 +83,14 @@
     },
     {
       "title": "Live Stream Archive",
-      "links": [{
-        "icon": "🔴",
-        "title": "Coding Train Live 186",
-        "url": "https://youtu.be/GMaKcRkiNoM?t=4054s",
-        "description": "Live Stream with Interactive Drawing with SketchRNN"
-      }]
+      "links": [
+        {
+          "icon": "🔴",
+          "title": "Coding Train Live 186",
+          "url": "https://youtu.be/GMaKcRkiNoM?t=4054s",
+          "description": "Live Stream with Interactive Drawing with SketchRNN"
+        }
+      ]
     }
   ]
 }
--- a/content/videos/ml5/0-introduction/1-introduction/index.json
+++ b/content/videos/ml5/0-introduction/1-introduction/index.json
@@ -48,7 +48,7 @@
         {
           "icon": "💻",
           "title": "ml5.js",
-          "url": "lhttps://ml5js.org/",
+          "url": "https://ml5js.org/",
           "description": "Reference material for the ml5 library."
         },
         {
--- a/content/videos/ml5/1-classification/1-image-classification/index.json
+++ b/content/videos/ml5/1-classification/1-image-classification/index.json
@@ -4,7 +4,7 @@
   "videoNumber": "1.1",
   "videoId": "yNkAuWz5lnY",
   "date": "2018-08-01",
-  "languages": ["ml5", " p5js", "javascript"],
+  "languages": ["ml5", "p5js", "javascript"],
   "topics": ["machine learning", "image classification", "MobileNet"],
   "canContribute": true,
   "timestamps": [
--- a/content/videos/ml5/2-transfer-learning/1-transfer-learning/index.json
+++ b/content/videos/ml5/2-transfer-learning/1-transfer-learning/index.json
@@ -5,7 +5,7 @@
   "videoId": "kRpZ5OqUY6Y",
   "date": "2018-08-13",
   "languages": ["ml5", "p5js"],
-  "topics": ["image classificaton", "feature extractor"],
+  "topics": ["image classification", "feature extractor"],
   "canContribute": true,
   "timestamps": [
     {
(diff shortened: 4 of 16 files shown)


shiffman:
Wow, this is incredible! Everything you are saying here makes sense on a high level to me and the code looks great! Before I merge let's see if @runemadsen, co-founder of DSI, might be able to review as well. #489 is absolutely in need of some attention if you are interested in working on it! 🚂💫

fturmel:
Great! I'm reading more on Gatsby (first time using it), and I am realizing that passing the list of topics and languages to the page context will create a bit of client-side duplication because this information will live both in the HTML markup and in JSON for React hydration. It's not a lot of data, but maybe there's a better approach there.

fturmel:
I saw a solid improvement locally with these last commits (`createPages` down to 11 seconds), but it's not making a dramatic difference during the Netlify builds 🤷

I think it's worth keeping, but we can also drop these changes. LMK!

shiffman:
Hi @fturmel I see this is still active with additional commits, could you let us know when this is ready to review? Thank you!! 💫

fturmel:
Sorry for all the noise here. I had an idea this morning I wanted to try out, and it spiraled into trying to troubleshoot incremental builds and cache related issues on the site. I still have a few things I want to look at, I'll let you know when it's ready and will keep it in draft mode until then. Hope that's alright!

fturmel:
@shiffman Thanks for your patience, I think we're in a great spot now and there's nothing more for me to change. I rebased and squashed all the noise and amended the initial PR description with additional info.

shiffman:
@fturmel wonderful, thank you so much for spending so much time on this!! I'll be glad to merge once @runemadsen is able to take a look! 🚀

fturmel:
@runemadsen Great, thanks for reviewing!

> the reason why the Netlify builds take longer is because the Netlify build servers don't have a lot of CPU [...]

For sure, weaker CPUs although I think they have 16 cores. The CI/CD also has more bookend work to do (load caches, setup, deploy to CDN, etc). When I wrote that comment, I was pondering about a specific commit during the PR process (squashed), I think related to the GraphQL regex query change. I saw something like an 8 seconds reduction for `createPages` locally, but it didn't move the needle at all during the Netlify build. Maybe it's a CPU architecture thing...

And to clarify, this PR _does_ improve the Netlify build and deploy numbers. They appear to be hovering between 6 and 8 minutes, with one (fluke?) even completing in 4 minutes. Most current builds are in the 10 to 12+ minutes range.

Revising this PR again this morning, and as one last step here we could remove the `Tag` nodes altogether. They aren't queried anymore and have been proven to become stale. It would be better to get rid of them to prevent their use in upcoming feature implementations. If that makes sense, I can push one more commit to wrap this up. Have a great day!

https://github.com/CodingTrain/thecodingtrain.com/blob/a9f9226f70cabc4a5014f817c92f2ac9af4d91a0/node-scripts/schema.js#L251-L254

https://github.com/CodingTrain/thecodingtrain.com/blob/a9f9226f70cabc4a5014f817c92f2ac9af4d91a0/node-scripts/node-generation.js#L144-L172

fturmel:
@shiffman Do you want to try to upgrade the Netlify plugin in the dashboard as well? It's the "Essential Gatsby" one https://app.netlify.com/plugins

Screenshot 2022-11-23 at 7 43 50 AM

shiffman:
Strange, according to the settings I am on the latest release?

Screen Shot 2022-11-23 at 9 35 28 AM

Am I missing another place to look?

fturmel:
Hrm.. I think this screen only displays the latest version number available and not necessarily the version installed. Plugins are installed at the project level, so you probably need to drill down a bit further. See what's available in the options dropdown.

shiffman:
Oh indeed, I was looking in the wrong place it's quite obvious, upgrading now!

runemadsen:
@fturmel I'm all for removing the `Tag` type if it's not being used!

fturmel:
Alright, good to go!

shiffman:
And merging! 🚀

PR #785, opened 2022-11-20
Update .nvmrc file

fturmel:
https://github.com/CodingTrain/thecodingtrain.com/blob/055797c74191d7a7c7f2cb3267bd9627b52948d4/CONTRIBUTING.md?plain=1#L52


shiffman:
Good catch!

PR #786, opened 2022-11-20
Use Netlify redirects instead of Gatsby's

fturmel:
https://thecodingtrain.com/discord still redirects to https://discord.com/invite/codingtrain even though it's been updated 2 days ago:

https://github.com/CodingTrain/thecodingtrain.com/blob/055797c74191d7a7c7f2cb3267bd9627b52948d4/redirects.json#L2

Maybe the cache could be purged by manually redeploying from the Netlify dashboard, but here I want to propose using Netlify's `_redirects` configuration file instead. It has a number of advantages and hopefully won't have the same cache issues if a URL needs to be updated in the future.

Gatsby's `createRedirect` generates a static HTML file that contains an HTML meta refresh tag, so there are no proper 3xx redirects happening. It's a bit slower and probably affects SEO negatively.

This `_redirects` file is at once more powerful and simpler to use (imo).
https://docs.netlify.com/routing/redirects/
https://docs.netlify.com/routing/redirects/redirect-options/

I also took the liberty of reorganizing by legacy/SEO links and vanity/shortcuts, the latter as 302 (temporary). This should prevent browsers from caching the redirect behavior, so these could potentially be updated during a live stream and "just work" 🤞


fturmel:
Works now: https://deploy-preview-786--codingtrain.netlify.app/discord

I had to add `!` at the end of each line in `_redirects` to give them priority over existing files/routes (the Gatsby generated HTML redirect pages are still in cache). Usually these would not be required.
https://docs.netlify.com/routing/redirects/rewrites-proxies/#shadowing

shiffman:
Thank you for this, it's great! I think the reason it might not have been working before is related to caching? I often notice the site not updating after merging a pull request. Only a force "clear cache and deploy site" seems to work.

fturmel:
> I often notice the site not updating after merging a pull request. Only a force "clear cache and deploy site" seems to work.

I think it might be the Gatsby cache that's responsible for this. I caught and fixed one of those cache issues while working on #781, and it was fully reproducible locally.

shiffman:
Oh interesting, thank you!

Issue #772, opened 2022-11-12
copy link button confirmation not visible when pushed to next line


dipamsen:
![image](https://user-images.githubusercontent.com/59444569/201487483-41906447-1af6-4d6e-babd-43fbbcd570a7.png)

When there are a lot of `groupLinks` entries, the additional entries overflows out of the screen instead of wrapping / scrolling.

Example: https://thecodingtrain.com/challenges/175-3D-cube-applesoft-basic

shiffman:
@kfahn22 and I were discussing this. I think maybe it's best for me to limit myself to 6 grouplinks max? For this one we can combine AppleSoft BASIC and Apple II Emulators. Maybe there's a more succinct way to say "unedited raw footage" as well.

kfahn22:
I think "raw" implies unedited.

dipamsen:
I guess we could technically add a horizontal scroll for this, but I'm not sure if it is good for the overall accessibility and design.

dipamsen:
Another point to note is, for the "Copy Link" button when wraps to the next line (which is the case for most recent challenges), on clicking on the button, the confirmation text "Copied Link!" appears on the left of the button, which is almost completely out of the screen.

shiffman:
I think it's good to just limit the number of tabs and length of text so it's not worth engineering a scroll. However, while low priority, it might be nice to fix the copy link button issue when the tabs push it to the next line.

Changed the title to reflect this new issue.

<img width="336" alt="Screen Shot 2022-11-14 at 10 29 33 AM" src="https://user-images.githubusercontent.com/191758/201699708-5315e929-0cd7-4916-a71f-77a161c7b425.png">

laurie60:
I think I fixed this issue, this is my first time contributing to an open source project!

laurie60:
Hello! I just realised that the my fix that was merged is not working unless the window is resized, so if you start off with a window size that gives you two lines, the alert still pops off screen. 

I have implemented a fix (to my fix!) by adding a useEffect that runs on the pages first rendering in another pull request #812.

I have tested it a lot now, using Chrome and Safari, and it works on my local build. Most of the time...

There is one situation where it still seems to fail though-which is on chrome where you navigate directly to a challenge video page. Where the alert should go to the right, it still goes to the left.  It does work on chrome when you navigate to the same page from another screen, and it also seems to work on Safari in all circumstances. I'm not sure why this is 🤔.

shiffman:
[...]

PR #789, opened 2022-11-24
fixed position of alert box

laurie60:
This solves the issue in issue #772.

### DIFF
--- a/src/components/ShareButton.js
+++ b/src/components/ShareButton.js
@@ -24,7 +24,12 @@ const CopyUrlToClipboard = ({ children, text, onCopy, options }) => {
   return cloneElement(elem, { onClick });
 };
 
-export const ShareButton = ({ className, variant, text = 'Copy link' }) => {
+export const ShareButton = ({
+  className,
+  variant,
+  wrapped,
+  text = 'Copy link'
+}) => {
   const [isCopied, setIsCopied] = useState(false);
 
   const onCopy = (text, res) => {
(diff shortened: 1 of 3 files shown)


shiffman:
Hi @laurie60! Congratulations on your first open source contribution! I am so thrilled to be that place for you! 

The issue is fixed when the button is on the line below the rest of the tabs. But now it looks like it is cutoff on a page where it not pushed to the next line!

Screen Shot 2022-11-24 at 10 04 22 AM

I wonder, is there a way it can detect where it is on the page? Aligned right or left?

Don't worry, I want to make sure we get this merged, I'll look into this too!

shiffman:
@all-contributors  add @laurie60 for code!

laurie60:
> Hi @laurie60! Congratulations on your first open source contribution! I am so thrilled to be that place for you!
> 
> The issue is fixed when the button is on the line below the rest of the tabs. But now it looks like it is cutoff on a page where it not pushed to the next line!
> 
> Screen Shot 2022-11-24 at 10 04 22 AM
> 
> I wonder, is there a way it can detect where it is on the page? Aligned right or left?
> 
> Don't worry, I want to make sure we get this merged, I'll look into this too!

Thank you! Oh flip, sorry I didn't notice that! One way I can think of solving this would be to use the screen width, as the copied button is only able to fit on the same line (where there are >= 5 tabs in the list) on quite wide screens.

But if there is a way to know the position of the button on the screen would be much more flexible.

dipamsen:
@laurie60 One possible solution could be to use a `ref` compare the top offsets (`elt.offsetTop`) of the button to check whether the button is on the same line or on the next line. 

From my experiments I found that when the button is on the right, the top offset is ~800, and when it is below on the left, the offset is ~850, across various zoom levels (that is on large screens, > 1180px. For medium screens, > 900px, this range becomes 850-900). 

If you have some other approach, please feel free to go your own way!

laurie60:
Thanks for the suggestion @dipamsen, I wasn't familiar with that concept, so I wouldn't have thought of it on my own. I implemented the useRef idea in the last commit, I think it is working as we want it to now!

shiffman:
Incredible! I can confirm it's working. Merging! Congratulations @laurie60 on your first open source contribution! ❤️

PR #796, opened 2022-11-25
Cleanup Tags

### LINKED ISSUE #488: Duplicate Languages Appearing in Dropdown
I noticed that there are duplicate languages appearing in the dropdown for different variations of the same language i.e, "processing" vs "Processing" and "p5js" vs. "p5.js"
Screen Shot 2022-08-24 at 7 38 10 PM
Screen Shot 2022-08-24 at 7 38 26 PM

This affects /tracks/lang/all/topic/all/ and /challenges. 

How do we want to normalize these languages so that there's one preferred way to refer to any given language?

dipamsen:
Removes duplicate language tags.

(Related: #488)

selector


shiffman:
Oh fantastic, can any of #489 be incorporated too or is that separate?

fturmel:
Interesting timing, I was just about to ask if I could pick up #489.  Should I still? I think it's very powerful to have a bulk normalizing / grouping / renaming pass when the Gatsby nodes get created for both topics and languages in addition to correcting the sources when possible.

shiffman:
Yes, I'd love to do both! I'm traveling today so can't easily review and merge things, but will be back online later tonight / tomorrow.

dipamsen:
@fturmel Just in case this is helpful, here are the original variations of `languages` which I have normalised (with the currently used versions at the beginning):
```
p5.js / p5js
ml5.js / ml5
Java / java
Processing / processing
JavaScript / javascript / Javascript
node.js / Node.js
```

fturmel:
@dipamsen Thanks for that!

shiffman:
I'm going to go ahead and merge this and we can tackle #489 next! @fturmel if you work on it would you mind forking that branch so that @M0nica's commits are included in the repo history? Thank you!!

PR #799, opened 2022-11-26
Normalize languages and topics (#489 continued)

fturmel:
Continuing #489 in this new PR.

WIP - I've only merged `main` into the branch, resolved conflicts, and undid a few leftovers from earlier false starts so far.


fturmel:
@shiffman Some questions:

If there are no replacements in the lookup map, the bulk normalization on topics and languages could either be:

1. untouched (ex: `bitwise AND`)
2. lowercased (ex: `bitwise and`)
3. lowercased + uppercase first character (ex: `Bitwise and`)

Currently it's option 3, which seems a little off as it's erasing a lot of proper casing in the sources. Option 1 and spot-fixing as needed with the lookup map seems like a better approach to me.

I was also thinking that it would make sense to merge some tags together when they have plural and singular wordings, again using the map. Ex: `array` and `arrays` would just be merged as `array` instead of having distinct entries.

fturmel:
@shiffman Another thing I am realizing now as I see the preview builds not reflecting the changes: we'll have more Gatsby cache issues with this approach because this transformation happens as we create the nodes.

If there is a Gatsby cache available and none of the source JSON files have changed, it won't bother trying to recreate the nodes... Changes to the lookup map would not get applied until the cache is cleared and the nodes are recreated.

I need to think about this a bit more, but one approach that could work is to transform the JSON source files themselves with a node script that runs just before the Gatsby build starts. Or run that script manually once in a while to cleanup the tags commit the changes to the repo.

shiffman:
> 



> If there are no replacements in the lookup map, the bulk normalization on topics and languages could either be:
> 
> 1. untouched (ex: `bitwise AND`)
> 2. lowercased (ex: `bitwise and`)
> 3. lowercased + uppercase first character (ex: `Bitwise and`)
> 
> Currently it's option 3, which seems a little off as it's erasing a lot of proper casing in the sources. Option 1 and spot-fixing as needed with the lookup map seems like a better approach to me.

I think I would prefer 1 or 2 yes! Let's either leave the casing or lowercase everything. What do you think?

> I was also thinking that it would make sense to merge some tags together when they have plural and singular wordings, again using the map. Ex: `array` and `arrays` would just be merged as `array` instead of having distinct entries.

Agreed!

shiffman:
> @shiffman Another thing I am realizing now as I see the preview builds not reflecting the changes: we'll have more Gatsby cache issues with this approach because this transformation happens as we create the nodes.
> 
> If there is a Gatsby cache available and none of the source JSON files have changed, it won't bother trying to recreate the nodes... Changes to the lookup map would not get applied until the cache is cleared and the nodes are recreated.
> 
> I need to think about this a bit more, but one approach that could work is to transform the JSON source files themselves with a node script that runs just before the Gatsby build starts. Or run that script manually once in a while to cleanup the tags and commit the changes to the repo.

Maybe it makes sense to file a separate issue and work on this in its own branch? This way we can merge the work on the topics. And for now it's pretty easy for me to run a "clear cache and build" command from netlify's admin dashboard.

fturmel:
@shiffman Almost done. It works great locally, just need to clean a few things up.

I've been doing a first content pass to try to make the topics clearer and more useful by themselves in the filter dropdown.

Some ideas that emerged, let me know if you agree:
- Capitalize correctly software, brands, people's names, etc
- Merge singular and plural together (ex: `arrays` and `array` => `array`)
- Suffix every game tag with the word `game`. Not everyone might know what `Plinko` or `Mad Libs` is.
- Tags that represent a p5.js function stylized as `.functionName()` (ex: `.beginShape()` instead of `beginShape`)
- I feel like we could drop the `creative coding` and `creative code` topics since they would apply to 99% of the content
- What do you think about merging `basics`, `for beginners`, `beginners`, `beginner` into one tag? I like `for beginners`
- `nature of code` tag only matches two challenges. Should we rename this as `Nature of Code book` and tag more content with this? (separate task)

shiffman:
This looks great to me! 

One general comment is that I think that more tags that are shorter / succinct is better. For example instead of having a tag be "Atari Asteroids Game" I would have 3 tags: `["Atari", "Asteroids", "Game"]`—what do you think?

I think we could also discuss this forever so I'm happy to merge this as is to get the framework up and running and we can tweak as necessary!

fturmel:
> One general comment is that I think that more tags that are shorter / succinct is better. For example instead of having a tag be "Atari Asteroids Game" I would have 3 tags: ["Atari", "Asteroids", "Game"]—what do you think?

Yeah that makes sense if there is enough content to tag for each. I think it's OK to repeat some words if it makes a tag clearer on its own. You could still have `["Atari", "Asteroids game", "games"]` for example. I liked that I could type "game" and actually see all the games right there in the topics, as well as a "games" umbrella term.

Screenshot 2022-11-29 at 8 02 03 AM

You're totally right though, it's easy to overanalyze this, and it can always be iterated upon. I'll pull this out of draft mode, let me know if you want to make any tweaks before merge.

About the implementation, as mentioned in a previous comment I did actually make the transformation happen with a nodejs script that modifies the source JSON files. That script runs before the build process on CI, and if you ever want to make these changes permanently to the source JSON, you can just run it and commit the changes. That approach also means no Gatsby cache issues.

fturmel:
Hold on - didn't see there was a unit test failing. It could be because the CI is still on node 14.x

One more thing, I did notice that tags that contain characters with special meaning for regex do not work correctly on the site. Ex: https://thecodingtrain.com/challenges/lang/all/topic/a*-algorithm

With this PR, some replacements use parens and it will make more tags return no results. I'll work on a fix for this today, I know what the problem is.

fturmel:
It was the GitHub actions nodejs version that was making the unit test fail. Fixed now.

Preview build: https://deploy-preview-799--codingtrain.netlify.app/

shiffman:
Awesome, this is great, merging!

PR #808, opened 2022-11-30
Fix tracks and challenges filtering for tags containing special regex characters

fturmel:
For example, the `A* algorithm` topic should return a result but doesn't: https://thecodingtrain.com/challenges/lang/all/topic/a*-algorithm

Preview build: https://deploy-preview-808--codingtrain.netlify.app/challenges/lang/all/topic/a*-algorithm


shiffman:
Thanks for catching this!

PR #809, opened 2022-11-30
fix double `and` typos

fturmel:


shiffman:
Thanks for catching these!

PR #812, opened 2022-12-03
fix so works without resizing window

laurie60:
This fixes the issue in issue 722, and should now work without needing to resize the window.


shiffman:
Amazing!

PR #805, opened 2022-11-29
Add link to PR in showcase submission confirmation

dipamsen:
image

Links to the generated pull request on GitHub along with the showcase submission confirmation.

(untested)

PR #823, opened 2022-12-08
Revert #805 (showcase PR preview URL)

dipamsen:
This PR reverts all changes in #805.

In #805, the github PR link was added on showcase submission, which allowed the contributor to go to the respective pull request for their showcase. However, it introduced an error due to which an error message was displayed upon submission, even though the showcase was being submitted and the pull request was being created. (initially found by @kfahn22)

I created and merged #822, where I removed the code which caused the bug to temporarily fix the issue.

As it turns out, the problem is that the showcase submission script is a Netlify Background function, and it is not possible to return any data from a background function to the client, thus preventing the client to receive the github PR number, which was being used to display its link to the contributor. 

So, it is not possible to link the contributor to their pull request, because the submission script is asynchronous in nature.

* [Background Function](https://docs.netlify.com/functions/background-functions/) docs on Netlify

----
PS: I found that the [`netlify` CLI](https://www.netlify.com/products/cli/) can be used to debug netlify functions locally, instead of creating creating new PRs every time.

### DIFF
--- a/netlify/functions/submission-background.js
+++ b/netlify/functions/submission-background.js
@@ -152,9 +152,4 @@ exports.handler = async function (event) {
   );
 
   console.log('Done!');
-
-  return {
-    statusCode: 200,
-    body: JSON.stringify({ success: true, pullRequestNo: prRes.data.number  }),
-  };
 };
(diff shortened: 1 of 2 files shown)


shiffman:
Thank you for this! We could include a generic link to https://github.com/CodingTrain/thecodingtrain.com/pulls so people know where to follow their submission?

shiffman:
Going to go ahead and merge this as well, thank you @dipamsen!

PR #818, opened 2022-12-07
[chore] Gatsby 5 migration, deps upgrade and minor fixes

fturmel:
Per #750, this upgrades Gatsby to v5 and most deps to their latest major versions. It should allow us to explore the new [Slices API](https://www.gatsbyjs.com/docs/how-to/performance/using-slices/) and keep improving build times.

Here's the official migration guide for reference: https://www.gatsbyjs.com/docs/reference/release-notes/migrating-from-v4-to-v5/

## Overview

- Upgrade to node 18, React 18, etc
- Rewrite GraphQL queries with new `sort` syntax
- Upgrade to MDX v2 - this was the majority of the work, a lot has changed
- Upgrade to `react-youtube` v10 - some props were renamed
- Remove `@mdx-js/mdx`, `html-react-parser` and `gatsby-plugin-meta-redirect`

And a few extras things:
- Add `trailingSlash: 'never'` in Gatsby config to get rid of 301s on initial page loads
  - This doesn't appear to work on the preview build, there could be additional things to configure on Netlify to make this work properly. Possibly related if we want to investigate: https://answers.netlify.com/t/remove-trailing-slash-redirect-for-gatsby-gatsby-cloud-netlify-website/20976/11
- Fix some React warnings: can't nest `p` tags, missing `tbody` in table, missing React keys in list
- Fix an event listener not being removed on component unmount

## Related work to consider?
Up for discussion imo, they could be done in separate PRs.

- [ ] Migrate from `react-helmet` to [Gatsby Head](https://www.gatsbyjs.com/docs/reference/built-in-components/gatsby-head/)
- [ ] Migrate all node scripts to ESM so we can use `remark-gfm` v3. See https://www.gatsbyjs.com/plugins/gatsby-plugin-mdx/#mdxoptions
  - Per https://github.com/gatsbyjs/gatsby/discussions/31599 there is a very recent effort to officially support ESM, maybe we should hold off for now?
- [ ] Find a way to migrate to latest `postcss-custom-media` and `postcss-custom-properties`? The `importFrom` option has been removed. See https://github.com/csstools/postcss-plugins/discussions/192

PR #826, opened 2022-12-09
Stabilize HTML build outputs

fturmel:
Some additional improvements to build/deploy performance.

It's been bothering me that the Netlify build often needs to redeploy thousands of files for no apparent reason after merging PRs that often don't change much. I've been diffing build outputs to understand why.

- the `react-select` component library increments element ids internally by default. This is causing a problem because Gatsby builds using parallel worker threads that execute in an unpredictable order, so a page will very likely end up with a different select component id between each build. Explicitly setting the `instanceId` prop fixes the problem.
https://github.com/JedWatson/react-select/blob/c350a124f8de853a79e0158578d438a35e67e63e/packages/react-select/src/Select.tsx#L638-L639

- the `TopBar` component that displays today's date is captured in HTML during the build phase. It creates some visual flickering because the date in the markup is often not the same as after hydration. It also forces every HTML page to be redeployed on a new build if the last build didn't happen on the same date (UTC). This PR only displays the date after React hydration to address the problem.

https://user-images.githubusercontent.com/4009209/206729522-5045becd-e87a-4277-a93e-ba62c40c0cb0.mp4

Screenshot 2022-12-09 at 9 44 41 AM

- Migrating to Gatsby 5(#818) seem to have helped a bit too, but there's one remaining issue I've been tracking down where the only thing changing between a clean build and the following incremental build is the `webpackCompilationHash` value. See https://github.com/gatsbyjs/gatsby/issues/33450. The fix was a breaking change for Gatsby 4 and was supposed to be rolled out in v5 - but I think it might've slipped. Not much we can do about this for now.

### DIFF
--- a/src/components/ItemsPage.js
+++ b/src/components/ItemsPage.js
@@ -129,6 +129,7 @@ const ItemsPage = ({
           selected={selectedLanguage}
           onChange={setSelectedLanguage}
           variant={variant}
+          instanceId="languages-filter"
         />
 
         <Select
@@ -140,6 +141,7 @@ const ItemsPage = ({
           selected={selectedTopic}
           onChange={setSelectedTopic}
           variant={variant}
+          instanceId="topics-filter"
         />
       </div>
 
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -14,7 +14,8 @@ export const Select = ({
   onChange,
   icon,
   className,
-  variant
+  variant,
+  instanceId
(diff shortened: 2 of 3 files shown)


shiffman:
I've checked this through and the deploy preview seems to be working great! Tagging @runemadsen just as an FYI, but I'm going to go ahead and merge!

runemadsen:
Lovely!

PR #831, opened 2022-12-17
Modified the url class in `--reduced` size in the `LinkList.module.css` file

zvolcsey:
Hi, I modified the url class in `--reduced` size in the `LinkList.module.css`. I opened the website in development mode and that seems to have fixed the problem.

PR #843, opened 2022-12-19
Fix link overflow on the right

### LINKED ISSUE #798: Group Link titles cut off on smaller screen sizes
In small screens, longer titles which wrap around on multiple lines are not visible properly.

image

zvolcsey:
Fix link overflow on the right side
Referring to this comment: https://github.com/CodingTrain/thecodingtrain.com/pull/831#issuecomment-1356137680

Should fix #798

### DIFF
--- a/src/components/LinkList.module.css
+++ b/src/components/LinkList.module.css
@@ -15,13 +15,9 @@
 .icon,
 .description,
 .url {
+  height: var(--baseline);
   padding: 0 var(--spacing-normal);
   line-height: var(--baseline);
-}
-
-.icon,
-.description {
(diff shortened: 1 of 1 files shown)


dipamsen:
By the way, I think it is fine if it overflows, after all that's the default behaviour of large screens! My comment was just for documentation purpose so that it would be easier to review and merge.

Either ways, thanks for working on this!

zvolcsey:
You are welcome! :) Currently long links look fine in smaller screen, when I look at the deploy preview.

shiffman:
Thanks for this @zvolcsey!

PR #845, opened 2022-12-20
`TopBar` and `Footer` component to Gatsby slices

fturmel:
I converted the `TopBar` and `Footer` component to use [Gatsby 5 slices](https://www.gatsbyjs.com/docs/how-to/performance/using-slices/). I'm not convinced we'll see a significant impact to build times here because that's not where the heavy lifting happens, but it can't hurt.

Eventually we could find other opportunities to use slices that could move the needle a bit more. For example, I was thinking the top part of the challenges page could be a good one to isolate and put in a slice because it's the same for at least 6k pages. And each content card will also appear in a lot of pages because of the filters and pagination, they can each be their own slice.

image

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -19,6 +19,7 @@ const {
   createAboutPageCoverImageNode
 } = require('./node-scripts/node-generation');
 const {
+  createSlices,
   createTrackVideoPages,
   createTracksPages,
   createChallengesPages,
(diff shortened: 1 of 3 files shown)


shiffman:
This is so great! @runemadsen, any thoughts here?

runemadsen:
I haven't worked with slices before, but it looks neat! I think it can be merged :)

shiffman:
Woohoo!!

PR #846, opened 2022-12-20
Fix character encoding for two showcase contributor names

fturmel:
There was something wrong with the character encoding in these JSON files.

Screenshot 2022-12-20 at 10 31 48 AM

Screenshot 2022-12-20 at 10 32 43 AM


shiffman:
❤️

PR #909, opened 2023-01-28
Add image validation to passenger submission lambda

fturmel:
Per recent attempts to upload backdoors/malware through the Passenger submission form, this is using the Sharp library to parse the submitted file in the lambda function and make sure it's an actual PNG or JPEG.

Based on the detected image format, it can determine the correct file extension to use, so we don't need `imageExt` anymore.

I also moved the GitHub token check up so it happens before it is passed to Octokit.

@shiffman @dipamsen Let me know what you think! I tested locally, but it should be possible to test it with the Netlify preview link. I didn't want to create test PRs that I couldn't delete myself.

### DIFF
--- a/netlify/functions/submission-background.js
+++ b/netlify/functions/submission-background.js
@@ -1,12 +1,12 @@
 const { Octokit } = require('@octokit/core');
 const slugify = require('slugify');
 const btoa = require('btoa');
+const sharp = require('sharp');
 
 // event.body expected to be:
 // {
 //   title: "Something",
 //   image: "base64string="
-//   imageExt: "png",
 //   authorName: "Coding Train",
 //   authorUrl: "https://thecodingtrain.com",
 //   authorEmail: "help@thecodingtrain.com",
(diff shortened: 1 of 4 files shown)


fturmel:
One more note: it would be possible to validate the file size and image dimensions here too, but I opted out of it. The lambda is a background process so we can't return validation errors to the client. It could've failed silently for legitimate submissions and cause some confusion.

dipamsen:
Confirming that this works!

Netlify Logs:

- Error:
error log
  

- Success:
success log

----

@shiffman An important detail to note is that as pointed out by @fturmel, this only results in a silent error, in both cases (valid/invalid submission), the following success message will be shown to the client.

image

Only if a valid image is submitted, will the pull request be generated on Github. However the client will not be alerted of whether their submission has gone through or not.

This is the limitation of using a background function, as no information (even HTTP status codes) can be sent back to the client. I am not sure if a normal function could be used instead, but leaving it out there as an option.

------

Suprisingly, the Netlify deploy preview works, even though it was "built without sensitive environment variables", `GITHUB_TOKEN` being one of them... ??

image

fturmel:
> This is the limitation of using a background function, as no information (even HTTP status codes) can be sent back to the client

Yeah, you get a 202 status code from the POST request which only confirms that it's been successfully queue for processing. It's probably possible to have sockets/polling to show the user when the background processing is done, but unlikely without a database and added complexity that I don't think is desirable here.

> I am not sure if a normal lambda function could be used instead, but leaving it out here as an option.

I could work, but I assume it was implemented this way because the max execution time for synchronous lambdas is 10 seconds (vs 15 minutes for async) and it's likely to time out when you consider all the GitHub operations taking place.

https://docs.netlify.com/functions/overview/#default-deployment-options

> Suprisingly, the Netlify deploy preview works [...]

Actually, this is a really valid point and a huge security vulnerability if any PR can log environment variables that contain infra secrets. In the Netlify settings, you can scope env variables so they are only available in prod. It's a fairly new feature.

https://www.netlify.com/blog/scopes-and-contextual-values-for-environment-variables-ga/

PR #915, opened 2023-01-31
Externalize train SVG in footer

fturmel:
I think we can get some benefits by not inlining this fairly large (60k, 15k gzipped) SVG in the footer of all pages.

Not sure yet if we'll notice a difference, but it certainly can't hurt imo!

- Most HTML pages generated are roughly 60k gzipped, so we are making the initial HTML payload 20% to 25% smaller. The file can be loaded later, it's below the fold in just about every scenario. And because it's an external file, it can be cached by the users browser.
- This _should_ make the site a bit faster to deploy on Netlify. 6686 static pages * 15KB is roughly 100MB less content to rsync (diff + network traffic).
- It will help reduce the DOM size.
Screenshot 2023-01-31 at 4 12 27 PM


shiffman:
Thanks! I'll merge and give this a try, we can always revert if there are unintended consequences!

PR #916, opened 2023-02-01
Fix anchors in homepage and guides

fturmel:
- Added missing `id` in homepage to make the "passenger showcase" footer anchor link work properly
- Replaced the `kebabCase` function with slugify in guides pages so that anchors work more predictably. It was transforming "GitHub" to "git-hub" for example and broke that specific anchor.


shiffman:
Amazing, thanks for catching this!

PR #920, opened 2023-02-03
Interactive multishaped starfield

Drea007:


shiffman:
Same comment as #922, the image file needs to have the same name as the json!

shiffman:
Looking good! 

Screen Shot 2023-02-03 at 7 44 11 PM

(will merge once the checks complete!)

shiffman:
Now live on the site!!!

PR #923, opened 2023-02-06
Add showcase timestamps (first batch)

dipamsen:
This PR adds the missing `submittedOn` fields for the showcase submissions from the new website form.

### DIFF
--- a/content/videos/challenges/116-lissajous-curve-table/showcase/contribution-1657913785.json
+++ b/content/videos/challenges/116-lissajous-curve-table/showcase/contribution-1657913785.json
@@ -3,5 +3,6 @@
   "author": {
     "name": "DukesTT"
   },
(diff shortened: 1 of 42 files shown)


fturmel:
@dipamsen awesome! I think we should make the `submittedOn` field required in the content test unit tests too.

https://github.com/CodingTrain/thecodingtrain.com/blob/339cda85a75d121e3650d33ca5088136bb4721ea/content-testing/file-formats.js#L172

shiffman:
This is incredible! How did you do it?

dipamsen:
@shiffman This is the first batch including all the new website form submissions (the old website ones are still remaining), it was the easiest because their filenames contain the timestamp 🙂

I am figuring out how I will go about automating the remaining ones.

@fturmel Yes, I will add that once all the files are done!

Again, sorry for not making it clear enough, this only includes some of the showcase submissions which were created by the website form. Next, I'll tackle the ones which were migrated.

PR #925, opened 2023-02-07
Fix video dates

fturmel:
While working on fixes to #638, I realized that some videos did not have dates (~68 of them). I wrote a little scraper to populate them with YouTube's published dates.

Then, I thought I could just run this on all videos to do a little content QA. It caught a few things, notably a bad videoId from the ml5 track.

Sorry for the sometimes larger diffs than required diffs, it seems like the JSON content files were not formatted using Prettier before.


fturmel:
I think that #870 could also probably be closed after this is merged. All missing NOC dates should've been added here.

shiffman:
incredible! Thank you!

PR #928, opened 2023-02-09
Showcase Submission for c4-worley-noise

ch00ooch00oo:
Thank you Rodak Titěra for your contribution! A member of the Coding Train team will review it shortly.

* [3D Worley Noise in Processing 4 using PShader](https://github.com/Rodak123/WorleyNoise)
* Rodak Titěra

preview image


shiffman:
Apologies it looks like there is a character that is breaking the JSON in your name, do you want to take a look or would you like me to try to adjust it?

fturmel:
@shiffman Pretty sure it's on our end while producing the JSON file from the lambda. The `ě` character in the author's name (Rodak Titěra) is not getting encoded in a JSON-compatible way. It would be nice to fix it for the contributor. I'll investigate the root cause.

shiffman:
Ah, thank you @fturmel! I manually edited the character so hopefully the checks will now pass as least as a band-aid here!

PR #929, opened 2023-02-09
Replace `btoa` with native Buffer base64 encoding

fturmel:
This should fix the encoding issue we've observed in #928


shiffman:
wonderful, thank you!

PR #934, opened 2023-02-12
Snake Game (rotting apple)

Drea007:
Get the code of my sketch from the Processing.org editor and run it in your Proccesing desktop app :)


dipamsen:
@Drea007 Looks like your image file is in the wrong folder!
image

The `contribution30.png` file should also be in the `showcase` directory, besides your `json` file.

shiffman:
Great work!

PR #943, opened 2023-02-21
remove `jest-extended`

fturmel:
In my opinion, this dependency is not bringing enough value to justify its use in the project. The few tests that were using extended `expect()` methods were easily rewritten with core ones.

`"bail": 0` in `package.json` could also be removed since it's already the default https://jestjs.io/docs/configuration#bail-number--boolean


shiffman:
Thanks for this @fturmel!

PR #944, opened 2023-02-21
Gatsby deferred static generation (DSG) tests

fturmel:
Trying this out on Netlify to see how it works and how much it can improve build/deploy times. Only the filtered pages would be deferred.

https://www.gatsbyjs.com/docs/conceptual/rendering-options/#deferred-static-generation-dsg
https://www.gatsbyjs.com/docs/how-to/rendering-options/using-deferred-static-generation/
https://www.gatsbyjs.com/docs/reference/rendering-options/deferred-static-generation/


fturmel:
Looks like it worked, there are 6263 DSG pages in the log. Build + deploy was pretty fast at roughly 3 min, but I expected more.

shiffman:
Awesome, thanks for taking a look at this @fturmel! Feel free to tag me when it's ready for review / merge!

fturmel:
@shiffman Looks like DSG doesn't work on the showcase pages only, it made the DSG lambda crash at runtime. I'm not sure why, maybe there is more info in the Netlify logs.

Screenshot 2023-02-21 at 6 52 45 PM

I suspect it's related to this "no function calls in gatsby-config file" limitation. `127` exit code is "command not found", it's possible that it was caused by a serialization problem. https://www.gatsbyjs.com/docs/reference/rendering-options/deferred-static-generation/#functions-inside-gatsby-config-are-not-allowed

It wasn't really improving the build times anyways, going to close this and take it as an interested exploration. This PR will at least document the attempt if we want to revisit DSG eventually.

fturmel:
Reopening this to eventually try against #949 (Gatsby 5.7) and see how it behaves. Just curious, it probably won't make any difference.

fturmel:
Nope! Same lambda crashes.

PR #946, opened 2023-02-23
Add timecodes for NOC ch3-5

kfahn22:
I have added the timecodes that were in the yt-description from 3-5.

### DIFF
--- a/content/videos/noc/3-angles/1-angles-and-rotation/index.json
+++ b/content/videos/noc/3-angles/1-angles-and-rotation/index.json
@@ -5,7 +5,18 @@
   "topics": ["For beginners"],
   "videoId": "DMg-WRfNB60",
   "canContribute": true,
-  "timestamps": [],
+  "timestamps": [
+    { "time": "0:00", "title": "Hello!" },
+    { "time": "0:25", "title": "Explain! Angles" },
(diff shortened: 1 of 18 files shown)


content/videos/noc/3-angles/1-angles-and-rotation/index.json
```diff
@@ -5,7 +5,18 @@
   "topics": ["For beginners"],
   "videoId": "DMg-WRfNB60",
   "canContribute": true,
-  "timestamps": [],
+  "timestamps": [
+    { "time": "0:00", "title": "Hello!" },
+    { "time": "0:25", "title": "Explain! Angles" },
```
shiffman, line 10: This was an idea we had a while back where we categorized timecodes as "Explain!" or "Code!" etc.  My thinking now is that perhaps this adds too much noise and maybe it makes more sense to keep the timecodes simple, short and sweet when possible. What do you think @kfahn22?

kfahn22: I just copied the ones from YouTube.  I agree it is probably unnecessary and  I will edit them.

kfahn22:
I have updated the timecodes.

shiffman:
Incredible thank you!

PR #948, opened 2023-02-24
Fix missing mouse pointer on "reset filters" button

fturmel:
(and just better use of the Button component in that context)


shiffman:
Excellent!

PR #949, opened 2023-02-24
Upgrade Gatsby 5.2 -> 5.7 + minor tweaks

fturmel:
Notably, 5.3 brought official ESM support for Gatsby config files, which means we can eventually use the latest `remark-gfm` version. Also version 5.5 introduced faster file hashing in the `gatsby-source-filesytem` plugin, we'll see if we notice any performance gains here.

Full release notes are here if you want more details: https://www.gatsbyjs.com/docs/reference/release-notes/

Bonus:

- Figured out a way to migrate to the latest postcss plugins. The removal of `importFrom` in `custom-media` and `custom-properties` was a blocker before, but the new [postcss-global-data](https://www.npmjs.com/package/@csstools/postcss-global-data) plugin works great as a replacement for this functionality. More context: https://github.com/csstools/postcss-plugins/issues/834

- All other dependencies bumped to latest (except `remark-gfm`, we'll get there)

- Changed the `siteUrl` metadata value in `gatsby-config` to "https://thecodingtrain.com" so it matches the actual domain in use and will stop 301 redirects for metadata urls.


shiffman:
Amazing, thank you!

Issue #524, opened 2022-09-03
Link to Nebula Video


shiffman:
I'd like to add to the JSON data for each video a nebula property that links to the video on nebula (if it is there). For example:

https://youtu.be/7r83N3c2kPw -> https://nebula.tv/videos/codingtrain-coding-challenge-173-applesoft-basic-snake-game

I'd love to then incorporate this into YT description generator (cc @dipamsen). We could also add it to the page for each video as well.

fturmel:
Working through adding the Nebula slugs to videos.

I extracted the titles and urls/slugs from the https://nebula.tv/codingtrain page (177 videos), and put them in this Google sheet to track progress: https://docs.google.com/spreadsheets/d/1-DpihoHjDcN2KEZufQBoxZ9zB-WblVjMzFIzYlFMGss/edit?usp=sharing

fturmel:
@shiffman Between #950 and #952 this covers all Nebula video I think. If you can review the spreadsheet, there are a few clarifications needed to complete this data entry task.

- The p5.js code track chapter 2 seems to have dupes/mismatches between the Coding Train vids and Nebula. Was chapter 2 re-recorded at some point? (these are highlighted in spreadsheet in yellow and cyan)
- Are there missing tracks on the Coding Train site still? I saw Nebula videos for "Teachable Machine" and "Discord bots" tracks.
- Some misc videos (like ITP show and uncut versions) I supposed are Nebula exclusive. I flagged them but it's probably expected that there are no matches on the site.
- I couldn't find a good match on the site for the Nebula video "Working with Data and APIs in JavaScript".

That's about it!

dipamsen:
> Are there missing tracks on the Coding Train site still? I saw Nebula videos for "Teachable Machine" and "Discord bots" tracks.
Quite a few... I think from the list in #279, the 'Learning Processing' videos are being remade, 'Discord Bots' and 'Twitter Bots' videos are pretty much outdated. The only one actually yet to be ported (from the list) is 'Guest Conductors'

fturmel:
@dipamsen thank you for your insights here! I updated one of the PR and the spreadsheet based on them.

There is a video on Nebula about JS objects (coding-train-2-3-javascript-objects-p5-js-tutorial) that doesn't have a match. Was it simply replaced by the 2.3 video (incrementation) and should be flagged as outdated?

shiffman:
Yes, this is an old video that I no longer include in the Code! track.

[...]

PR #950, opened 2023-02-25
Add `nebulaSlug` to challenge videos

fturmel:
some progress on #524


fturmel:
@shiffman Quick note, challenge 54 has 2 parts, so I added `nebulaSlug` as a valid property inside the parts array.

fturmel:
@shiffman both Nebula PRs can be merged

PR #956, opened 2023-02-27
Add code of conduct and license

dipamsen:
Added code of conduct (from [here](https://github.com/CodingTrain/website-archive/blob/main/CODE_OF_CONDUCT.md)) and the MIT License.


shiffman:
Thank you @dipamsen! Does it make more sense to link to https://github.com/CodingTrain/Code-of-Conduct or duplicate it here?

Also, I'd like to add a note to the README along the lines of:

> Please note that while the code in this repository is licensed under the MIT License, any brand and visual assets included in this repository are not covered by the license. These assets, including but not limited to images, icons, and logos, are solely intended for use on thecodingtrain.com and should not be reused or distributed without express permission. If you have any questions regarding the use of these assets, please contact help@thecodingtrain.com.

What do you think?

shiffman:
Noting this would cover #575 #574

dipamsen:
> Does it make more sense to link to https://github.com/CodingTrain/Code-of-Conduct or duplicate it here?

The only reason I like it being duplicated is that GitHub directly shows a link to the Code Of Conduct (and license) at the top of the repo, instead of having to scroll down to the bottom of the README.

> Also, I'd like to add a note to the README along the lines of [...]

Great idea!

shiffman:
> The only reason I like it being duplicated is that GitHub directly shows a link to the Code Of Conduct (and license) at the top of the repo, instead of having to scroll down to the bottom of the README. But definitely both are possible.

That makes sense! Maybe we should deprecate the other repo and have this be the main doc? It's been a while since I revisited the Code of Conduct and it might make sense to consider any updates / changes.

shiffman:
I archived and made a note in https://github.com/CodingTrain/Code-of-Conduct

PR #960, opened 2023-03-02
link to Nebula videos from challenges and tracks

fturmel:
As discussed in #524, took a stab at implementing the Nebula links. If a `nebulaSlug` is present in a video or video part, it'll add an extra row below the video with a link out.

Screenshot 2023-03-02 at 1 17 06 PM


shiffman:
This is great, thank you!! I've spoken with Nebula and they are going to help design a custom button which we can add next week. Stay tuned!

fturmel:
@shiffman Latest commits restore the collapsible single row when we don't have Nebula slugs, and implement the new button design.

shiffman:
Looks fantastic, thank you!!

PR #961, opened 2023-03-02
Upgrade deps (glob 9.x, remark-gfm 3.x) + gatsby-config ESM

fturmel:
This brings all project deps to their latest versions.

As a reminder, we needed Gatsby 5.3 and the `gatsby-config` file to be ESM to be able to upgrade to `remark-gfm` 3.x.

Could be worth migrating the other Gatsby nodejs files to ESM eventually, maybe even every nodejs code in the repo.


shiffman:
So appreciate you taking the time to update these!!!

fturmel:
No problem!

PR #963, opened 2023-03-04
Add copy to clipboard in yt-desc (migrate to ESM)

dipamsen:
Added copy to clipboard option to `yt-desc`

```
npm run yt-desc path/to/index.json -- -c
```

(Using `-c` directly doesn't work, because for using `npm run` only passes option args if prepended with `--`.)

Also migrated to ESM because `clipboardy@3` only supports ESM.


shiffman:
This is awesome, thank you!

PR #985, opened 2023-03-30
Update homepage 3-30

kfahn22:
Updated the featured challenges and showcases.
Added Buffon's needle to Pi Day track
Changed featured challenge to 52-random walker


shiffman:
Thank you!

PR #987, opened 2023-03-31
More topics cleanup

shiffman:
I think I got everything related to "fractal" now!


fturmel:
btw you could've added an entry for "fractals" in node-scripts/tags-transforms.json for this!

shiffman:
Ah, right, I forgot about that, I'm going to add it now!

PR #989, opened 2023-04-02
Fractal Tag Transform

shiffman:
Thanks to suggestion from @fturmel #987


fturmel:
Shouldn't the key just be "fractal", unless you meant to capitalize the output of the transformation?

shiffman:
> Shouldn't the key just be "fractal", unless you meant to capitalize the output of the transformation?

Yes, you are right! Updating...

PR #993, opened 2023-04-06
Cleanup tags and Lint content files

dipamsen:
1. ran `tags-transform` script to standardise video tags

```bash
npm run tags-transforms
```

2. ran `prettier` linter to all content json files

```bash
npx prettier --write "./content/**/*.json"
```


shiffman:
This is great, thank you! I've often wondered if there are particular settings I should adopt in VSCode so that when I "format on save" for any files in this repo it matches the prettier conventions of the codebase here?

PR #1003, opened 2023-04-15
Adding physics libraries side track - [WIP]

kfahn22:
Adding Matter.js and ToxicLibs physics tutorials to NOC track.

I have linked to two of the code examples from NOC ch. 6 (for 5.15 and 5.16).  I am not sure if they need to be edited. Also no links to processing sketches yet.


shiffman:
This is incredible @kfahn22! I think that I don't want to include older Java syntax videos as part of this playlist, however, I would love to have those as a side track. What do you think about taking out the Toxiclibs videos from NOC here (maybe I'll make new ones eventually) and instead putting them in a side track. The matter ones can stay. We can also add the Soft Body Coding Challenge to the NOC track after the matter.js videos.

kfahn22:
Yes, I agree.  Especially since you are using the js version of Toxiclibs  in Ch. 6.

kfahn22:
I have created a new pull request only adding only the Matter.js videos (and Soft Body) to NOC.   Will leave this open and edit to add a Physics library track.

dipamsen:
@kfahn22 after the completion of your PR, you could run `npm run formatc` to style the files edited according to linting rules defined by the repo.

Issue #1005, opened 2023-04-15
Missing `relatedChallenges`


fturmel:
If someone wants to complete this data entry task, I found only 11 challenges on the site that didn't have related challenges.

- [x] 148-gift-wrapping
- [x] 157-zoom-annotations
- [x] 164-slitscan
- [x] 166-image-to-ascii
- [x] 167-prime-spiral
- [x] 170-monty-hall-problem
- [x] 171-wave-function-collapse
- [x] 172-directional-boring
- [x] 56-attraction-and-repulsion-forces
- [x] 67-pong
- [x] c3-hilbert-curve

kfahn22:
I have started working on this.

fturmel:
Re-opening this, `c3-hilbert-curve` is left

shiffman:
#1015 covers `c3-hilbert-curve` so I believe this can be closed?

fturmel:
Yes, we can close this

PR #1006, opened 2023-04-16
additional content JSON validation tests + content fixes

fturmel:
Finished up a proof of concept to use `yup` and stricter checks to catch more content issues.

There is some overlap with `format.test.js` though, so let me know how you feel about this. Maybe the approach can be better integrated into the current system, like using the `yup` schemas in `content-structure.js` and `file-formats.js`.


dipamsen:
This is amazing! Some comments:

The `relatedChallenges` field is actually valid for all videos, not just challenges. (even though they are currently not being used in other videos.) Also, related challenges can be empty, for videos which aren't directly related to a challenge, eg. beginner tutorials.

(See: https://github.com/CodingTrain/thecodingtrain.com/issues/262#issuecomment-1154381530)

From `content-structure-guide.md`:
> **Related challenges**
> [...] If it's not defined or is left empty, then no challenge panel is shown in the corresponding video page. [...]

fturmel:
For sure, we can put back `relatedChallenges` for non-challenge videos.

I was trying to see how strict the validation can be against the data that currently exists. It looked to me like the content was headed towards always having related challenges in challenge videos (see #1005), while never having them in others.

dipamsen:
Sure, you're right that the content has currently been tending in this direction.

@shiffman can provide more insights on this - should the related challenges field be used for non challenge videos?

shiffman:
> @shiffman can provide more insights on this - should the related challenges field be used for non challenge videos (Currently none of them have any...)

Yes, the original intention of the site was to have the option for related challenges on *all* videos, challenges or not. For example, if there is a p5.js intro video about arrays maybe we would put the Particle System coding challenge as related? There are also lots of related challenges for the NOC videos. I just never got to the point of expanding this feature out. I'd like to keep it but open to the community's input of course!

fturmel:
@shiffman with #1007 merged, how do you feel about requiring at least one related challenge for all challenges?

shiffman:
I'm ok with that! My only concern is that sometimes I'm in a rush to get a webpage up for a new video and it can be convenient to skip it and add later. But if you think enforcing the requirement is good, I'm 100% on board!

fturmel:
> I'm ok with that! My only concern is that sometimes I'm in a rush to get a webpage up for a new video and it can be convenient to skip it and add later. But if you think enforcing the requirement is good, I'm 100% on board!

Honestly it's up to you. If you value the flexibility more that's perfectly fine. In my experience, the "add it later" part usually gets forgotten. As I mentioned previously in this PR, I just noticed the tendencies to want all challenges to have related challenges and if that's the goal then adding a guardrail is an improvement imo.

shiffman:
Allright, let's try it! We can always change it later. I'd like to keep the option of relatedChallenges for regular videos, maybe we should open an issue to have someone specifically work on filling those out around the site.

fturmel:
Tests are currently failing because the `c3-hilbert-curve` challenge is the only one without related challenges. I reopened #1005

shiffman:
Amazing, super excited to merge this in, thank you for the incredible efforts @fturmel!

fturmel:
You're welcome, glad you like it!

PR #1012, opened 2023-04-22
Add submittedOn for challenge 1 and 2

dipamsen:
## Added `submittedOn` field for showcase - challenges 1 and 2

I am using the git blames from the old repo ([example](https://github.com/CodingTrain/website-archive/blame/main/_CodingChallenges/002-mengersponge.md)) to locate the date of PR/submission of old challenges.

I don't know if this can be automated, as there is no identifier (id) for a showcase submission, by which we can relate them from the old site to the new site (possibly `url`), also there isn't an API to get the Blame information as far as I know.

<sub>(2 down, 167 more to go...)</sub>


dipamsen:
While merging this, please use `Squash and Merge`, so not keep the WIP commits.

shiffman:
So this is good to go? I noticed a lot of other changes here I'm assuming they are just whitespace / code formatting adjustments? Thank you for taking a look at this, it's awesome!

fturmel:
@dipamsen Heads up, this PR is removing the `source` prop in many contribution files.

dipamsen:
@shiffman yes, just includes linting on the soft body file. (also this is c-1 and 2 only, 3 (snake) has like 30 showcases lol)

dipamsen:
> @dipamsen Heads up, this PR is removing the `source` prop in many contribution files.

the `source` prop is deprecated and is not used on the site, it was only kept in the json files for archival purposes, but is not really required

fturmel:
> @fturmel [the `source` prop is deprecated](https://github.com/CodingTrain/thecodingtrain.com/issues/244#issuecomment-1153171288) and is not used on the site, it was only kept in the json files for archival purposes, but is not really required

I didn't realize that, thanks for linking the source!

shiffman:
Awesome!! (And yes, I think we can remove `source` since we can always look it on old submissions on the archived website archive if we needed to access it.)

PR #1016, opened 2023-04-23
changing nebula slug for climate spiral

shiffman:


shiffman:
This was covered in #1017

PR #1021, opened 2023-04-25
Separating showcase submission function into synchronous and background

dipamsen:
Netlify background functions have a maximum payload size limit of 256 kB, which is very small for sending all the data for a showcase, and can cause unexpected errors. 

This PR breaks down the submission function into two separate functions - one synchronous and one background. 

Initially, the client sends all the data to the synchronous function, which uploads the contribution image as a blob to GitHub, and passes the rest of the data to the background function. The background function then takes over and does the rest of the jobs. This allows us to have a total of 6 MB limit for all the contribution data, as well as having a maximum execution time of 15 minutes (in the background function), as opposed to a only 10 seconds.

This PR also adds a cryptographic signature validation between these two functions, using the `GITHUB_TOKEN` variable. This is required, because the background function on its own is publicly available (at `/.netlify/functions/submission-background`), so we need to make sure that it does not handle any other requests, other than those originated from the synchronous function itself.

---
## Breakdown of tasks

**Synchronous Function**
- Create Image Blob
- Sign the payload

**Background Function**
- Verify signed payload
- Create JSON Blob
- Create a Git Tree with the Image and JSON
- Create a commit on `main`
- Create a new branch pointing to the commit
- Create a PR against `main`
- Add `showcase` label

----
- [x] Use GitHub Database API
- [x] Validation/Security 
- [x] Use HTTP header to pass the signature instead of in the body itself


shiffman:
Hey @dipamsen this looks fantastic! I just tried it via the preview deployment, should it work there or it won't until merged in main? I got an error for a test submission.

fturmel:
You could also verify the error in the Netlify logs to make sure the error really is related to the token not being available?

shiffman:
Indeed I can confirm the logs say "Github Token not found", I'll merge and run a test, we can always roll back if there is an issue. Thank you @dipamsen and @fturmel I am so excited for this one!

PR #178, opened 2022-06-07
fixed link to passenger showcase guide in the footer component

Gabrz:

PR #1025, opened 2023-04-27
Created a new showcase submission in cc #178

raj-jaiswal:
Manually submitted a showcase in coding challenge #178 since the automatic submission from the website is not working. I was not sure about the number at the end of the file name so please check that.


raj-jaiswal:
Here's the contribution image

fturmel:
Thanks for submitting directly through GitHub! I think your image was just a _bit_ too large to get through the form submission. #1021 should address the 256 KB limit once completed.

raj-jaiswal:
Thanks for the reply! I do think there should be a warning about file size when it gets too big. All it showed me was that it could not be submitted and to contact at the help email.

shiffman:
Thank you for this! Yes, hopefully we'll have the issue resolved soon thank you for the submission and feedback @raj-jaiswal!

PR #1030, opened 2023-05-01
Fix showcases filtering edge case (issue #1029)

### LINKED ISSUE #1029: Showcase Page - Different cases of same author causes unexpected behaviour
When there are multiple submissions with more than one casing of their name, then the author slug formed are the same for them, and only a single casing dominates in the showcase page.

Example: 
image

image
(Both the links link to the same page)

image
(But only one of their contribution is listed)

fturmel:
Closes #1029.

This PR adds a new `author.nameSlug` resolver on the `Author` Gatsby node, which is then used by the showcases page to correctly filter by _slugified_ author name.

Screenshot 2023-05-01 at 12 48 54 PM

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -28,6 +28,7 @@ const {
   createShowcasePages
 } = require('./node-scripts/page-generation');
 const set = require('lodash/set');
+const { toSlug } = require('./node-scripts/utils');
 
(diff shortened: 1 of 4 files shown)


shiffman:
Fantastic! I'm going to merge this and then go and manually fix the errors you found too.

PR #1034, opened 2023-05-02
fix: add dependency node-fetch

dipamsen:
Continuation of #1021

Added dependency `node-fetch@2` since node 14 does not support fetch.


shiffman:
Upgraded to node 18 so no longer needed!

PR #1038, opened 2023-05-03
Misc Cleanup in anticipation of video about the website

shiffman:
I am making a video about the new website and cleaning up small details I'm finding as we prep to capture the pages.


shiffman:
Website capturing is happening today so merging this!

PR #1041, opened 2023-05-03
Fix resolver that finds tags for tracks containing challenges

### LINKED ISSUE #1039: Track topics and languages
I need to investigate this more but I noticed that several of the side tracks do not show topics or languages. I believe these are meant to be compiled from the topics and languages of the videos that are contained within the tracks? I see the following three scenarios:

* all topics/languages compiled
* only topics/languages from the first video in the track
* no topics/languages listed

I'm not sure what causes one of the above scenarios to occur? Does anyone have an idea?

Example:
Screen Shot 2023-05-03 at 2 50 54 PM

fturmel:
Closes #1039

The track resolver that locates tags was not using the Challenge nodes correctly.

### DIFF
--- a/gatsby-node.js
+++ b/gatsby-node.js
@@ -267,11 +267,8 @@ const tagResolver = async (source, context, type) => {
     }
   }
 
-  // fetch all video nodes we found
-  const allVideos = await context.nodeModel.getNodesByIds({
-    ids: videoIds,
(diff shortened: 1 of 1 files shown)


dipamsen:
Seems to work! 
image

shiffman:
Fantastic!! Thank you!!

PR #1043, opened 2023-05-03
Minor tags UX improvements

fturmel:
While we're fixing tags, here are a few adjustments that I had noted down but never spent the time to implement.

- Fix the expand/collapse toggle behavior. The "show less" option was never properly displaying.

- Don't display "show more" when there's only 1 more tag to show. The button itself takes up the space it would've taken to display the remaining tag.

- Don't show the last trailing comma if we have a "show more" button

- Remove padding on the "show more" button so it aligns better when it's first in a row

- Display the amount of tags to be revealed in the "show more" button

The track `ml5js-beginners-guide` illustrates these changes well:

**Before**
https://thecodingtrain.com/tracks/ml5js-beginners-guide

https://user-images.githubusercontent.com/4009209/236052090-fd163510-5665-4d09-8d35-8f77f17faaa8.mp4

**After**
https://deploy-preview-1043--codingtrain.netlify.app/tracks/ml5js-beginners-guide

https://user-images.githubusercontent.com/4009209/236052095-27cf9946-8ae9-4526-9e61-1cd8875a6ba5.mp4

### DIFF
--- a/src/components/Tags.js
+++ b/src/components/Tags.js
@@ -13,12 +13,12 @@ const Tags = memo(
     linkTo,
     headerType = 'h2'
   }) => {
-    const [showAll, setShowAll] = useState(false);
-    const truncatedSize = 2;
-    const overflowSize = items.length - truncatedSize;
-    const allowToggle = overflowSize > 1;
+    const [truncate, setTruncate] = useState(true);
+    const truncateSize = 2;
+    const overflowSize = items.length - truncateSize;
+    const showToggle = overflowSize > 1;
     const visibleItems =
-      !allowToggle || showAll ? items : items.slice(0, truncatedSize);
+      showToggle && truncate ? items.slice(0, truncateSize) : items;
 
     const Header = headerType;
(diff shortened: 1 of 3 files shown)


fturmel:
One other option to consider: change the copy on the "show more" button so it reads more fluidly:

p5.js, ml5.js, <ins>and 2 more</ins>

shiffman:
> One other option to consider: change the copy on the "show more" button so it flows better. 
> p5.js, ml5.js, and 2 more

Oh! I like this! If you have a preference, feel free to go with your instincts!

fturmel:
Yeah, I think I like this approach too. I lowered the `margin-right` in the `tracks` page as well, it looked off.

Screenshot 2023-05-03 at 10 21 54 PM

Screenshot 2023-05-03 at 10 22 33 PM

Screenshot 2023-05-03 at 10 22 51 PM

shiffman:
Awesome, merging!

PR #1044, opened 2023-05-05
Page CTA - More Customisations

dipamsen:
Allows to customise character, and variant of the CTA sections of every page:

image

image

image

image

(Changed showcase page cta color and character)


shiffman:
@jasontheillustrator any thoughts on these?

dipamsen:
I am unable to fix the error atm, if this is required right now, someone can push the fix - the fix is to either set Character={some character} in <PagePanel> in components.js, or set default value for the prop in Page Panel itself.

(I can probably do it myself within a few hours)

fturmel:
I can handle it, give me a few minutes

shiffman:
Amazing, thank you @fturmel and @dipamsen!

jasontheillustrator:
> @jasontheillustrator any thoughts on these?

these look good to me.

PR #1045, opened 2023-05-09
fix Gatsby cache problem related to tracks

fturmel:
The digest for track Chapter nodes was computed using the `data` variable, which previously had its `videos` property removed.

https://github.com/fturmel/thecodingtrain.com/blob/c2d6f21e5b6ebd64e9c7cbca8d75ab4b285ad44c/node-scripts/node-generation.js#L185C10-L203


shiffman:
Ah, great catch, thank you!!

PR #1049, opened 2023-05-11
Add showcase to polar curves

kfahn22:


dipamsen:
Btw, the author.url property should probably be a link to the user instead of repo.

PR #1053, opened 2023-05-13
Gatsby scripts refactor to ESM + tracks order cache fix

fturmel:
It's a fairly big rewrite, but a notable code quality improvement imo.

- Moved all Gatsby scripts to `./gatsby` directory, reorganized for better code co-location (less imports)
- Converted all Gatsby scripts to ESM, and generally modernizing the code
- Re-implemented the `gatsby-awesome-pagination` plugin to ~35 lines of internal code for ESM compat and because the plugin doesn't appear to be maintained anymore
- Rewrote a few utility functions for clarity
- Removed references to guest tutorials and talks elements in content-testing and GraphQL schema (legacy / never implemented?)
- Fixed yet another Gatsby cache issue, this time related to tracks order
- Reduced Gatsby node creation code by approximately 50% (~500 LOC) by introducing a `createDefaults` helper function and using a routing pattern based on plugin and instance name.

Hope you like it!


gatsby/createResolvers.mjs
```diff
@@ -0,0 +1,229 @@
...
+    // normalize to ISO 8601 since some dates are in `yyyy-mm-dd` format
+    return new Date(source.submittedOn).toISOString();
+  } else {
+    // the contribution doesn't have a `submittedOn` property, let's derive one for sorting purposes
+
+    // extract the sequential number from the source JSON filename
+    const [seqNumber] = source.name.match(/\d+/g);
+
+    // edge case where the sequential number is actually a timestamp yet `submittedOn` was not set
+    if (seqNumber.length > 6) return new Date(+seqNumber).toISOString();
```
dipamsen, line 158: afaik this edge case shouldn't be necessary as all such videos already have their `submittedOn` set

fturmel: You're right, I verified and we don't have this edge case in the content anymore. I believe we did when the resolver was originally written but we can probably remove this. Thanks for the review!

shiffman:
This is fantastic, excited to merge, thanks for such a thoughtful and comprehensive update for the site!

PR #1056, opened 2023-05-15
Remove duplicate contributions

fturmel:
These look like test entries or mistakes? They are in the NoC welcome video (no contributions flag set) and the same submission was already correctly submitted in the [NoC random vectors video](https://thecodingtrain.com/tracks/the-nature-of-code-2/noc/1-vectors/3-random-vectors).


shiffman:
Yup, I can confirm this same contribution is part of the random vector video, thanks for finding this, merging!

PR #1058, opened 2023-05-15
Add submittedOn field to challenge showcase

dipamsen:


fturmel:
Let me know if you need help!

Probably parsing the `git blame` output and keeping the oldest date per contribution object is good enough of a strategy to automate this.

image

dipamsen:
@fturmel that's probably the best option. There might be some incorrect data due to files being updated later, but I think it is good enough.

fturmel:
The `author:` line is very unlikely to have been modified after the initial commit though, unless some whitespace was changed.

Another thing I just noticed: in some cases, [the original commit was in a README.md](https://github.com/CodingTrain/website-archive/blame/34524b57bd580753c4bb9ea0fa95e63f10837c83/CodingChallenges/CC_003_Snake_game_p5.js/README.md) file and got moved during a content restructure. It appears to have been done in [this single PR](https://github.com/CodingTrain/website-archive/pull/822). Maybe you could keep a list of these commit hashes, and as you parse the `git blame` output you could flag a contribution if any of its lines contains one of the hashes so they can be processed separately.

shiffman:
@dipamsen is this ready for final review / merge? I love that you are doing this but also if it becomes really onerous or tedious I don't think it's a high priority for the site to have perfect date sorting for the original showcase from the older site.

dipamsen:
@shiffman gtg 🚀

PR #1059, opened 2023-05-16
Adding images for hillbert curve and marching squares

kfahn22:
Cont. 13 in marching squares is not loading for me.  

Deleting two showcases that  are no longer available.


fturmel:
re: Cont. 13 in marching squares

These `ddns.net` urls are commonly used to self-host on a dynamic IP, so it's very likely that this link will remain dead.

I searched for "Metaballs" in the user's GitHub though and did find the source code if we want to link to that instead: https://github.com/YushchenkoAndrew/template/tree/master/JS/Metaballs

fturmel:
The same user had another contribution hosted on the same domain if you want to change both at once: `content/videos/challenges/113-4d-hypercube-aka-tesseract/showcase/contribution6.json`

kfahn22:
Thanks I updated the links.  It is a shame his website is no longer available.

shiffman:
Thank you @kfahn22!

PR #1063, opened 2023-05-18
Add submission dates to challenge showcases

dipamsen:
Based on the git blame output of the old website repo, this PR automatically fills the submittedOn property on the contributions (challenges only).

[Update: I was also able to get the data for cabana videos]

Out of 1303 contributions, 1065 (~81%) could be automatically filled. 90 (7%) of them had to be manually filled due to moving of data / unable to match the contribution from old to new website / the contribution only existing after the new website.
Rest of the contributions already have the field.

In  this PR, all challenge submission dates are added.


shiffman:
Thanks again @dipamsen!

PR #1073, opened 2023-05-23
Update Heroku-hosted contribution links

fturmel:
Not surprising, but since they ended their free tier, none of the Heroku hosted contributions were working anymore. I did some hunting and found alternate links or GitHub sources for most of them.


shiffman:
This is so helpful, thank you @fturmel!

PR #1074, opened 2023-05-24
Update showcase (climate spiral)

dipamsen:
Using curves to draw the climate spiral to be closer to the original visualisation.

image


fturmel:
I think my recent Gatsby refactor introduced a subtle bug that is causing this build to fail. When an existing image gets updated, the CoverImage node is still pointing to the previous File id which doesn't exist anymore. Working on finding a solution and properly identify the root cause, but in the meantime clearing the Gatsby cache works.

fturmel:
@shiffman as mentioned in #1075 it should be the Gatsby cache on Netlify. Once the `main@HEAD` build finishes, we could try to just re-run the checks & deploy on this PR. I think it might be enough to fix things.

shiffman:
I just forced a rebuild of this branch with "clear cache" via netlify's interface, maybe that will do the trick!

fturmel:
Yes. Something is off though, the checks UI on GitHub seems stuck and didn't run the unit tests.

Screenshot 2023-05-30 at 10 15 18 AM

fturmel:
I'm having the same problem on #1086 yet as you pointed out #1084 went through. Am I the one making everything break today somehow? 😂

fturmel:
Should be safe to merge locally.... OH hold on something just unblocked suddenly! Maybe merging #1084 helped

shiffman:
Finally! Great job as usual @dipamsen

PR #1075, opened 2023-05-24
Fix incremental Gatsby builds when an existing cover image changed

fturmel:
This is something I introduced in #1053 (sorry!).

For some reason, if we copy the `File` node properties when creating a `CoverImage` node, it will cause the `childImageSharp` property to be missing only on incremental builds and when the image source was modified.

You can see how it created a build failure on #1074

We may require a full Gatsby cache clear along with this build so that all `CoverImage` nodes get recreated with the required properties only.


shiffman:
Looks great, thank you for catching this!

fturmel:
@shiffman don't forget to clear the Netlify cache!

> We may require a full Gatsby cache clear along with this PR merge so that all CoverImage nodes get recreated with the required properties only.

shiffman:
Yes, it's happening as we speak, thank you!!

PR #1082, opened 2023-05-30
Update showcase climate spiral (retry)

fturmel:
This PR is just #1074 but coming from my GitHub account. Hoping this can fix the GitHub UI being stuck on the original branch.


fturmel:
So strange... unit tests did not run here either and we can merge the PR from the UI.

PR #1092, opened 2023-06-02
Showcase Submission for 71-minesweeper

ch00ooch00oo:
Thank you Ray Bellis for your contribution! A member of the Coding Train team will review it shortly.

* [Minesweeper with more features](https://editor.p5js.org/raybellis/sketches/p1EFR6BS5)
* Ray Bellis

preview image


fturmel:
@shiffman I think I know what's going on... The cache on the `main` branch wasn't cleared after merging #1075, only the caches for the PR branches. When a new PR comes in, it inherits the `main` cache and keeps failing with the same error we saw for #1074.

If we clear the cache on `main` and for this PR, I think we'll be in the clear going forward.

shiffman:
@fturmel ah, that makes sense! Let me see if I can clear these caches now.

shiffman:
Main branch was cleared and rebuilt now i'm running this one. When it completes I may need to make a trivial change to get GitHub actions to run again.

shiffman:
Wow, the animation is super impressive!

PR #1094, opened 2023-06-02
Custom sort order for guides

fturmel:
I think they currently are displayed in the order that the Gatsby MDX nodes are being created, which ends up being different for every build as this is an async process.

This PR sorts using the frontmatter date value, newest first. Unfortunately, we can't sort in the GraphQL query directly because of the `mm/dd/yyyy` format and need to parse them as native `Date` objects first.


shiffman:
I think this might something I would like to customize rather than sort by date? What do you think about creating a `GuideOrder` property much like we have for `tracks` in [`content/tracks/index.json`](https://github.com/CodingTrain/thecodingtrain.com/blob/main/content/tracks/index.json)?

fturmel:
Sure, we can do that too

fturmel:
@shiffman Should be good to go. I added a `guidesOrder` property to `content/pages/guides/index.json` and kept the sorting client-side for simplicity.

There are also new content tests for guides and tracks order so that they always contain explicit references to all guides and tracks.

shiffman:
Amazing, this is wonderful, thank you @fturmel!

PR #1095, opened 2023-06-02
Improve content testing errors

fturmel:
By setting the `abortEarly` option to `false` in the `yup` library and restructuring the unit tests a bit, we can show all schema validation failures (not just the first) and display the Jest errors in a more readable way.

**Before**
Screenshot 2023-06-02 at 6 49 53 PM

**After**
Screenshot 2023-06-02 at 6 45 00 PM


shiffman:
Wow, this will so incredibly helpful!!

Issue #26, opened 2021-10-27
Tracks page


fdoflorenzano:
As of 27/10, this page component was started by Rune. It holds most of the structure present in the designs in the unfiltered state. There expanded version of track cards are missing, and the logic associated to actually filtering the results based on a selected filter.

- [x] Page header
- [x] New to coding CTA
- [x] Character pop out
- [x] Filter panels
- [x] Base Track Card
- [x] Expanded Track Card
- [x] Filtering logic
- [x] Mobile version

Some tweaks to make:
- [x] The breadcrumbs should be removed, since it doesn't really provide useful navigation at this point
- [x] New to Coding CTA should collapse in mobile 
- [x] Borders of filter panels look weird when expanded

PR #1110, opened 2023-06-08
flexii showcase submission 3d supershape

flexiiwiki:
#26 3d supershapes coding challenge showcase submission


shiffman:
Thanks for sharing!

PR #1112, opened 2023-06-10
Add sound classifier video to ml5 track.

kfahn22:
I have also  added the 147-chrome dinosaur challenge to the track.

### DIFF
--- a/content/tracks/main-tracks/ml5js-beginners-guide/index.json
+++ b/content/tracks/main-tracks/ml5js-beginners-guide/index.json
@@ -13,7 +13,8 @@
       "videos": [
         "ml5/1-classification/1-image-classification",
         "ml5/1-classification/2-webcam-classification",
-        "ml5/1-classification/3-object-detection"
+        "ml5/1-classification/3-object-detection",
+        "ml5/9-doodleNet/1-doodleNet"
       ]
     },
     {
@@ -57,10 +58,6 @@
         "ml5/8-convolutional-neural-network/3-max-pooling",
         "ml5/8-convolutional-neural-network/4-image-classification"
       ]
-    },
-    {
-      "title": "DoodleNet",
-      "videos": ["ml5/9-doodleNet/1-doodleNet"]
     },
     {
       "title": "Sound Classification",
--- a/content/tracks/main-tracks/ml5js-beginners-guide/index.json
+++ b/content/tracks/main-tracks/ml5js-beginners-guide/index.json
@@ -61,6 +61,10 @@
     {
       "title": "DoodleNet",
       "videos": ["ml5/9-doodleNet/1-doodleNet"]
-    }
+    },
+     {
+       "title": "Sound Classification",
+       "videos": ["ml5/10-sound-classification/1-sound-classification"]
+     }
   ]
 }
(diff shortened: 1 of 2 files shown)


shiffman:
@kfahn22 i'm not sure if this makes sense but I moved doodlenet to the image classification section. What do you think?

kfahn22:
That makes sense-it was a single video by itself.  BTW, I was thinking of adding 158-shape classifier to the track and tried the sketches, butI couldn't get them to load in Safari.  I did try on Google chrome and worked as expected.  Maybe it is just me?   (The doodlenet sketches work for me in Safari.)

dipamsen:
Wondering if "ml5.js: " in the video titles are redundant?

image

shiffman:
> That makes sense-it was a single video by itself. BTW, I was thinking of adding 158-shape classifier to the track and tried the sketches, butI couldn't get them to load in Safari. I did try on Google chrome and worked as expected. Maybe it is just me? (The doodlenet sketches work for me in Safari.)

My guess is that you have WEBGL disabled in safari? It's working for me in safari. Although that's odd if doodlenet is fine?

shiffman:
> Wondering if "ml5.js: " in the video titles are redundant?
> 

Yeah, this is a good point. The titles pull from YouTube and it's helpful to have it in there for SEO. But I could move it to the end or I'm open to a way to override titles on the website if we want?

Train your own Neural Network with ml5.js

kfahn22:
I have a lot of sketches using WEBGL, so I don't think that is it.  It is possible it is something to do with my laptop, though.

dipamsen:
> > Wondering if "ml5.js: " in the video titles are redundant?
> 
> Yeah, this is a good point. The titles pull from YouTube and it's helpful to have it in there for SEO. But I could move it to the end or I'm open to a way to override titles on the website if we want?
> 
> Train your own Neural Network with ml5.js

Website titles are totally independent from youtube titles, so it should be fine, right? I think  on the website, there is enough context to the viewer that the video is using ml5.js, unlike youtube where it can be randomly recommended to them.

shiffman:
> Website titles are totally independent from youtube titles, so it should be fine, right? I think on the website, there is enough context to the viewer that the video is using ml5.js, unlike youtube where it can be randomly recommended to them.

Oh, you are right, I was confused! Yes, let's remove ml5 from them for sure, it's redundant! Also, it looks nicer when the titles fit on one line so might be worth keeping that in mind when shortening. Feel free to open a new branch to work on this if you like!

kfahn22:
I have MacOS 13.3 and Safari 16.4, so that is consistent.

dipamsen:
> I agree the ml5 is unnecessary.  Fixing depends on how much effort it would take-- I would love to know a simple approach!  Sorry to ask if this is obvious, but how is the SEO done for the website--I don't see a SEO file in components.

I think just remove the initial ml5.js in the titles of these videos (in `index.json`). The SEO for website should be good, since the video pages themselves do have a title of the track ml5.js (and i believe the same is true for the meta tags for SEO) so it is not required in the video title.

kfahn22:
I was actually wondering if there was a more efficient approach to removing the "ml5.js".

dipamsen:
@kfahn22 Probably a quick find and replace on the ml5 folder would work

dipamsen:
To answer your question on seo, the SEO stuff is actually set in `components/Head.js` - the actual content is derived from the content of the page, eg. The `title` will be title of the challenge/track/video on that page, etc.

kfahn22:
Thanks @dipamsen!

kfahn22:
Sorry, I didn't intend to open two pull requests.  Should I close this one?

PR #1126, opened 2023-06-22
Geometry showcase

kfahn22:
Adding an example of how to use p5.Geometry to create a custom shape.


shiffman:
This is fantastic, thank you for sharing! When I get done with NOC I'd love to do a video about p5.Geometry.

PR #1129, opened 2023-06-26
Refactor ShareButton + minor improvements

fturmel:
A collection of very minor improvements while poking around the site's share button functionality.

- The `ShareButton` implementation looked odd and convoluted to me. I cleaned it up a bit, removed the `copy-to-clipboard` dependency and just used the native Clipboard API which has been [well supported for some time now](https://caniuse.com/async-clipboard).
- Fixed a "copied" [notification position bug](https://github.com/CodingTrain/thecodingtrain.com/issues/772#issuecomment-1335975436) on page load, and also refactored the `Tabs` component so it's a bit more readable and idiomatic (refs over DOM selector).
- Converted a few divs with `role="button"` into actual buttons so that keyboard events don't have to be re-implemented. `onKeyPress` has been [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event) anyways.
- Standardized import name of the `classnames` package to `cn` everywhere (only two components were different)
- Vertically aligned some grid lines when the share button wraps to the next line:

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14,7 +14,6 @@
         "@octokit/core": "^4.2.0",
         "classnames": "^2.3.2",
         "clipboardy": "^3.0.0",
-        "copy-to-clipboard": "^3.3.3",
         "gatsby": "^5.7.0",
         "gatsby-core-utils": "^4.7.0",
         "gatsby-plugin-image": "^3.7.0",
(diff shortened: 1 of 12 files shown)


shiffman:
Thanks for all these adjustments, looks great!

PR #1130, opened 2023-06-26
Showcase form - all videos

dipamsen:
Adds support for submitting to non-challenge videos from the Passenger Showcase Form!

image

**Details:**

- Only videos with `"canContribute": true` are listed on the page.
- Videos which are present in multiple tracks, are only listed under their `canonicalTrack`. (eg. the track "Algorithmic Botany" is not listed at all; as all its videos fall under "Coding Challenges")
- A "Coding Challenges" track is added manually to the beginning of the tracks list, so it is listed along with other tracks. I thought this made the most sense in incorporating track videos and challenge videos together.
(This is done in the `useVideosWithShowcase` custom hook)
- The order of list of tracks on the page is arbitrary (as returned by GraphQL), except for Coding Challenges which is always at  the beginning of the list.
- The order of the videos on the page is based on the actual video ordering in the track; except for videos in Coding Challenges which are ordered by Date: Newest to Oldest.

## Deploy Preview
🚂 [Preview](https://deploy-preview-1130--codingtrain.netlify.app/guides/passenger-showcase-guide)

## Testing
Tested through [Netlify CLI](https://www.netlify.com/products/cli/)
See #1131 and #1132

## Screenshots

image

image

image


dipamsen:
Open to improving the copy for description of the fields "Track" and "Video" - any suggestions?
image

src/components/PassengerShowcaseForm.js
```diff
@@ -56,11 +54,92 @@ const PassengerShowcaseForm = () => {
           videoNumber
         }
       }
+      tracks: allTrack {
+        edges {
```
fturmel, line 58: You could flatten this to just use `nodes` instead of `edges` and `node`.

src/components/PassengerShowcaseForm.js
```diff
@@ -56,11 +54,92 @@ const PassengerShowcaseForm = () => {
...
+      }
     }
   `);
 
+  const tracks = data.tracks.edges
+    .map((edge) => edge.node)
+    .map((track) => {
+      // gather all videos from chapters (main track)
+      if (!track.videos && track.chapters) {
+        track.videos = track.chapters.reduce((acc, chapter) => {
```
fturmel, line 92: `flatMap` instead of `reduce` would be more concise.

src/components/PassengerShowcaseForm.js
```diff
@@ -56,11 +54,92 @@ const PassengerShowcaseForm = () => {
...
+        }, []);
+      }
+      delete track.chapters;
+      return track;
+    })
+    .map((track) => {
+      // filter out videos that can't be contributed to
+      track.videos = track.videos.filter((video) => video.canContribute);
+      // keep only videos that belong to this track
+      track.videos = track.videos.filter(
```
fturmel, line 103: Maybe you could chain the two filters and assign to `track.videos` only once? I initially thought it was a mistake and the second filter was undoing the first. Not a big deal.

src/components/PassengerShowcaseForm.js
```diff
@@ -56,11 +54,92 @@ const PassengerShowcaseForm = () => {
...
+    title: 'Coding Challenges',
+    slug: 'challenges',
+    videos: data.challenges.nodes.map((node) => {
+      return {
+        title: `#${node.videoNumber} ${node.title}`,
+        slug: node.slug
+      };
+    })
+  };
+  return [challengesTrack, ...tracks];
```
fturmel, line 121: It would be a nice optimization to memoize this whole transform and filtering operation. The hook will run every time a form value is updated yet always returns the same value.

fturmel (commented):
Awesome, really good stuff! I left a few minor notes to consider.

src/components/PassengerShowcaseForm.js
dipamsen: I thought about this but couldn't find a good way to do this. Is it maybe a good idea to save the output in a state variable on mount?

fturmel: Yeah, useEffect + useState could work, but memo would be even simpler. Inside the hook itself, try something like this:

```js
const data = useStaticQuery // ...

return useMemo(()=>{ /* your filter code */ }, [data]);
```

dipamsen: That makes sense, thanks for your input!

shiffman:
So excited to finally have this as part of the site!

PR #1139, opened 2023-06-29
Auto-select the track and video submission form fields when possible

fturmel:
Had this idea yesterday, works really well!

We're simply passing a state object using the router when the user clicks on one of the "submit to the showcase" CTAs in a track or challenge. This allows the form to know where the user came from and select the correct `track` and `video` values for them.

https://github.com/CodingTrain/thecodingtrain.com/assets/4009209/01e698cf-22dc-4413-82bc-4cb22ff4c73b

### DIFF
--- a/src/components/Button.js
+++ b/src/components/Button.js
@@ -13,6 +13,7 @@ export const Button = ({
   children,
   disabled,
   rainbow,
+  state,
   ...otherProps
 }) => {
   const classes = cn(css.root, className, {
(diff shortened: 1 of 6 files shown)


dipamsen (approved):
Amazing!

shiffman:
Wow, this is incredible, thank you @fturmel! I'll definitely be showing this off when I'm back to streaming after finishing Nature of Code writing!

fturmel:
Glad you like it!

PR #1152, opened 2023-07-13
Cleanup 

dipamsen:
- rename "Noise Track" -> "Noise"
- move video `open-simplex-noise` inside `noise/` folder
- add `"10 PRINT"` topic to challenge video


shiffman:
Fantastic, thank you so much for this!

PR #1157, opened 2023-07-20
display credits for Track videos

fturmel:
@dipamsen pointed out on Discord that the Track videos credits were not being displayed, even though they exist (and are required) in the source JSON files. This should fix the issue.


shiffman:
Thank you @dipamsen and @fturmel!

PR #1160, opened 2023-07-22
Cleanup `memo` usage

fturmel:
This is a very minor cleanup around the usage of React `memo()`. It removes a few duplicate calls and tries to be more consistent with its usage by applying the memoization along with the `export` statement when possible.


shiffman:
Amazing!

PR #1163, opened 2023-07-23
[Showcase Form] Group track videos into chapters

dipamsen:
🚂 [Deploy Preview](https://deploy-preview-1163--codingtrain.netlify.app/guides/passenger-showcase-guide)

Groups videos according to chapters

Normal tracks work as usual

<sub>(I don't think this is as important but thought that it would be cool to have)</sub>

### DIFF
--- a/src/components/PassengerShowcaseForm.js
+++ b/src/components/PassengerShowcaseForm.js
@@ -88,24 +88,22 @@ const useVideosWithShowcase = function () {
   return useMemo(() => {
     const tracks = data.tracks.nodes
       .map((track) => {
-        // gather all videos from chapters (main track)
-        if (!track.videos && track.chapters) {
-          track.videos = track.chapters.flatMap((chapter) => chapter.videos);
-        }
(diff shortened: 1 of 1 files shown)


src/components/PassengerShowcaseForm.js
```diff
@@ -238,9 +238,24 @@ const PassengerShowcaseForm = () => {
           Video
           <select name="video" value={state.video} onChange={onChange}>
             <option value="">Select a video</option>
-            {(
-              data.find((track) => track.slug === state.track)?.videos || []
-            ).map((node) => {
+
+            {(selectedTrack?.chapters || []).map((node) => (
```
fturmel, line 242: You could just continue the optional chain here, a little nicer than the empty array imo. Same on line 258.

`selectedTrack?.chapters?.map(...)`

fturmel (commented):
Solid improvement, really like it!

shiffman:
This is fantastic, wow!

PR #1180, opened 2023-08-20
Create showcase-contibitution - shapes and line

Mehakdoracode:
Adding my first showcase edition and also my first code .

PR #1186, opened 2023-08-28
Revert "Create showcase-contibitution - shapes and line"

shiffman:
Reverts CodingTrain/thecodingtrain.com#1180


shiffman:
Ah, yes, re-opening is better!

PR #1188, opened 2023-08-30
Create random

Mehakdoracode:
adding my showcase submission!


kfahn22:
Hi @Mehakdoracode!  Here are a couple of things you should update so the pull request will pass the tests. 

The file is in json format and must end in .json.  There is already a file in the showcase folder named contribution1.json, so I recommend you change the name of your showcase to contribution2.json.   If you did that, the path would be:

content/videos/code/2-variables/4-random/showcase/contribution2.json

 You can look in the folder to see how the other showcases are named to get an idea of what it should look like.  

The url field and date field need to be updated as follows:

-the url field should be:  

"url": "https://www.linkedin.com/in/mehak-dora-aaa596241"

-the date field must be formatted like this:

"submittedOn": "2023-08-30T6:55:08.000Z"

Ideally, you should also upload an image from your sketch. You can grab a screen shot or download an image.  ( In case you don't know how to do this, the following code will download an image for you.)  The image should have the same name as the .json file.  So if you change the name to "contribution2.json", the image would be named "contribution2.jpg" and should be in the showcase folder. Note that you might need to edit the image because the images must be a maximum of 800 pixels wide.

function mousePressed() {
 save('contribution2.jpg');
}

If you would like help with any of these things, please let me know and I would be happy to assist you!

Mehakdoracode:
Thank you, I will do that.

shiffman:
Thank you @kfahn22 for helping with this! Great work @Mehakdoracode!

fturmel:
Should be fixed now

shiffman:
Thank you @fturmel!

PR #1198, opened 2023-09-03
Add Neural network track

kfahn22:
This is ready for review.  I only added links to the GitHub repos with the finished libraries for the matrix videos and feedforward/backpropagation videos.  Please let me know if there are any other code examples that should be added.

I didn't link to one book you recommended (Neural Networks and Deep Learning) b/c the site is  http and there is a suggested donation (so people might enter credit card info).  

I have also alphabetized the side tracks.  I think it makes it easier to find a track and current order does not reflect any objective criteria.  Maybe you have a preference re the order--please let me know, but current order is somewhat arbitrary.


shiffman:
Wow! Amazing! I'll merge and @jasontheillustrator when you have a chance we can add a new thumbnail for this track.

See: https://deploy-preview-1198--codingtrain.netlify.app/tracks/neural-networks

Also, the noise track could use one! I may have missed adding it to the list before.

Screen Shot 2023-09-04 at 11 02 47 AM

PR #1217, opened 2023-09-17
Final backfilling submittedOn

dipamsen:
Backfilled ALL submittedOn values!

@fturmel are there any content tests to be updated after this change? I have set `isRequired` in the basic file format tests.

### DIFF
--- a/content-testing/file-formats.js
+++ b/content-testing/file-formats.js
@@ -172,7 +172,7 @@ const contribution = {
         }
       },
       url: { isRequired: true, type: 'string' },
-      submittedOn: { type: 'string' }
+      submittedOn: { isRequired: true, type: 'string' }
     }
   }
 };
--- a/content-testing/format-augmented.test.js
+++ b/content-testing/format-augmented.test.js
@@ -322,7 +322,7 @@ const challengesSchema = baseVideosSchema.concat(
 const showcasesSchema = strictObject({
   title: string().required(),
   url: string().url().required(),
-  submittedOn: dateRangeValidator, // TODO should be required, but we need to backfill these dates before enforcing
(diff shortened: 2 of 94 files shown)


fturmel:
@dipamsen great work!

You can make the field required here: https://github.com/CodingTrain/thecodingtrain.com/blob/ffc1f8565309cc3510a8de0ce92fadd212b6697a/content-testing/format-augmented.test.js#L325

You could also greatly simplify the logic in the resolver:

https://github.com/CodingTrain/thecodingtrain.com/blob/ffc1f8565309cc3510a8de0ce92fadd212b6697a/gatsby/createResolvers.mjs#L145-L167

You're reminding me that I've been sitting on some updates to the content tests for months... I'll create a draft PR today before I completely forget about it!

fturmel:
You might be able to get rid of the resolver completely. We wouldn't depend on a value found in another node anymore (the referenced video's date), so it eliminates the potential cache issue that justified the resolver to begin with.

The line that normalizes to a full UTC ISO 8601 datetime could go in the node creation phase and be dealt with at build time:

https://github.com/CodingTrain/thecodingtrain.com/blob/ffc1f8565309cc3510a8de0ce92fadd212b6697a/gatsby/onCreateNode.mjs#L18-L37

fturmel:
Nice! One last minor thing, you can remove the comment that `submittedOn` is implemented as a resolver in the schema:

https://github.com/CodingTrain/thecodingtrain.com/blob/eb5f64ad5779ffe997d486f28ff9818c44089cda/gatsby/createSchemaCustomization.mjs#L97-L98C24

fturmel (approved):

PR #1220, opened 2023-09-20
Adding reference to paul merrell's Model Synthesis

shiffman:


dipamsen:
@shiffman does the "more info below" comment make sense for the website? (The description and links are shown on separate tabs)

PR #1214, opened 2023-09-16
Learning Processing

dipamsen:
- [x] Cover Image

image

image

image

🚂 [Deploy Preview](https://deploy-preview-1214--codingtrain.netlify.app/tracks/learning-processing)

PR #1230, opened 2023-10-01
Don't pluralize video count label for tracks with only one video

fturmel:
New edge case since tracks always contained multiple videos until #1214


shiffman:
Sorry, missed this one, thank you for fixing @fturmel!!

PR #1237, opened 2023-10-19
Invertible cube showcase

MMMMMNG:
Two or so years ago I was inspired by this invertible cube model to try and code up a simulation of its movement.
I have been a fan of the coding train for quite some time, which is how I knew p5.js. 
So I used that.
I used p5js WEBGL rendering and matrix math (with the library math.js) to implement the invertible cube animation, and as it turned out, that worked really well! Even though I thought matrices were quite a challenging concept, they were super useful for reasoning about the transformations I had to perform.

Recently I saw the video on CodingTrain about Hacktoberfest, and so I thought "now is the perfect time to finally touch up and show off this project" so this is it.

Thanks CodingTrain for being so awesome!
Dan, you rule! 💝
Cheers
-MNG


shiffman:
This is fantastic, thank you for sharing!

PR #1246, opened 2023-10-22
Improved content tests

fturmel:
This is wrapping up what was started in #1006. I've been working on this off and on for a little while now, but just got to finishing it!

We now have yup schemas with slug reference checks for videos, challenges, showcases, FAQs, tracks, the guides page, homepage, FAQs page, tracks index and about page. I think that's the whole site except for the 404 page?

Error messages were improved. A few validators were added, like the timestamps array validator which makes sure that video timestamps are sequential.

The track type directory prefix isn't required anymore in `content/tracks/index.json`. I realized that because the URLs are structured as `/tracks/{slug}`, these slugs actually _have_ to be unique or something will break. A new test has been added to ensure uniqueness between side and main tracks.

With apologies to @fdoflorenzano and @runemadsen, I stripped out the original format validation since the new approach now validates more content with a stricter approach. Doubling the errors reported and slowing the content test completion didn't make much sense. The content tests for file presence and directory structure are still in place though!

Hope you like it!

EDIT: added a schema for the "about" page after all


dipamsen (approved):
Great stuff!

content-testing/schemas.js
```diff
@@ -0,0 +1,343 @@
...
+    })
+  ),
+
+  groupLinks: array(
+    strictObject({
+      title: string().required(),
+      links: array(
+        strictObject({
+          title: string().required(),
+          url: urlOrRelativeLinkValidator.required(), // TODO ref check instead of regex pattern matching for internal links?
```
dipamsen, line 108: This might be a good idea, the yt-desc script kind of checks for this, as it tries to convert all relative links to youtube links. Though here we'd need to check for other kinds of links rather than just video/track pages.

https://github.com/CodingTrain/thecodingtrain.com/blob/00545fd2621645b508780c5d14a7f8df5ac3bc5a/node-scripts/yt-description.mjs#L186-L228

Here's a bad url I just found:
https://github.com/CodingTrain/thecodingtrain.com/blob/00545fd2621645b508780c5d14a7f8df5ac3bc5a/content/videos/webgl/6-createGraphics/index.json#L63-L64

fturmel: That was aspirational when I initially wrote this comment, but yeah! We already have all the site slugs in sets, it's probably pretty easy at this point to make a new validator that would handle this. I'll take a stab at it later today, thanks for reminding me.

fturmel: Done! Caught a second bad link in there.

dipamsen (approved):
Looks good! Just one comment - the /showcase path can also take subpaths (for filter by author), but we probably won't link there anyways, so it is fine.

content-testing/validators.js
```diff
@@ -0,0 +1,285 @@
...
+
+  // sequential timestamps
+  let errors = [];
+  let previousTime = -1;
+
+  for (const value of values) {
+    const currentTime = timestampToSeconds(value.time);
+    if (currentTime <= previousTime) errors.push(value);
+    previousTime = currentTime;
+  }
```
dipamsen, line 218: Another possible test may be to check if the first timestamp is `0:00` (or `00:00`/`0:00:00`), or else youtube doesn't parse the timestamps as chapters.

fturmel: Yes, I had coded this at some point but removed it for some reason. Maybe thinking it was too strict? I did not realize it would break YouTube chapters, sounds like it should be added back in.

fturmel:
> Looks good! Just one comment - the /showcase path can also take subpaths (for filter by author), but we probably won't link there anyways, so it is fine.

I left out filters, pagination, and URI fragments. Don't think we need to bother since they are not used anywhere, but I can add a reminder in the code base to that effect.

content-testing/validators.js
fturmel: I corrected 5 videos that were missing `0:00` timestamp, you can see them in [94e0c03](https://github.com/CodingTrain/thecodingtrain.com/pull/1246/commits/94e0c03750ab90a1f3b292f89b8e776d05067076)

fturmel: I also found documentation on [YouTube video chapters](https://support.google.com/youtube/answer/9884579?hl=en), we can also validate that there are at least 3 timestamps per video (if any), and that they are at least 10 seconds between each other.

fturmel:
PR in draft mode now. Tests will fail - the last timestamp validation rules I added caught 16 issues (chapters too close from each other and videos with only one or two timestamp). Let me know how you'd like to resolve these.

dipamsen:
I have fixed the timestamps, but due to some reason I can't push my changes on here :/

dipamsen:
I did it!

fturmel:
@dipamsen Thanks for the help!

dipamsen:
@shiffman this is ready for review/merge! 
(I am waiting on this to be merged to work on #1252)

shiffman:
This is incredible work! I did a cursory scan of it and I am so impressed by the care and thoughtfulness you both put into this work. I am eternally grateful!

PR #1247, opened 2023-10-22
add chapters to Physics side track

### LINKED ISSUE #1098: Should "Physics Libraries" be a main track?
It makes sense to me that the videos in this track can be separated into chapters - Matter.js, toxiclibs, and Box2D

image

Even the actual files of the track are separated into folders:
image

which creates urls, suggesting a main-track like structure:
```
https://thecodingtrain.com/tracks/physics-libraries/physics-libraries/toxiclibs/3-connected-systems
```

On the other hand, this track doesn't completely fit the definition of a main track: 
> “Main Tracks” are sequenced video tutorials that you can follow like a course syllabus.

It is a bit closer to what a side track is (except the order part):
> “Side Tracks” are collections of related videos but don’t necessarily need to be watched in order.

It is somewhere in between...

Thoughts?

fturmel:
Closes #1098, I realized no one had actually contributed the proposed change.

### DIFF
--- a/content/tracks/side-tracks/physics-libraries/index.json
+++ b/content/tracks/side-tracks/physics-libraries/index.json
@@ -1,32 +1,47 @@
 {
   "title": "Physics Libraries",
-  "description": "Tutorials on the Matter.js, Toxiclibs, and Box2d physics libraries",
-  "videos": [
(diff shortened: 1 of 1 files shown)


shiffman:
Amazing, thank you!

Issue #1252, opened 2023-10-25
Correction Timecodes


shiffman:
YouTube now offers corrections timecodes, I just tried adding one for this video:

https://thecodingtrain.com/tracks/pixels/pixels/painting-with-pixels

I manually added it to the YT description:

```
Corrections: 
02:40 This old example is not using classes in JavaScript. Check the accompanying code for an updated version!
```

<img width="435" alt="Screen Shot 2023-10-25 at 11 40 31 AM" src="https://github.com/CodingTrain/thecodingtrain.com/assets/191758/269792f3-d988-493f-bae6-0c7f87a64137">

Maybe we should consider adding this to the JSON for each video as well as the description generator?

dipamsen:
Would this need to be implemented on the website as well? YT doesn't seem to show the pop-up inside embeds.

shiffman:
> Would this need to be implemented on the website as well? YT doesn't seem to show the pop-up inside embeds.

I don't think it needs to be no, but since I generate the descriptions from the JSON, I'd love to have it there for safe-keeping and consistency.We could render it to another "tab" like references, videos, etc. "Corrections" also. 

This is pretty low priority as I only just used it for the first time on this one video... and don't know if I really plan on adding it to more places in the future. In theory, though, it's a great feature to be able to annotate mistakes and things that have changed.

fturmel:
Can we close this issue since #1282 was merged?

shiffman:
Yes! Thanks for the reminder!

PR #1282, opened 2023-11-04
Add corrections + enhancement for yt-desc

dipamsen:
## Corrections

Add correction timestamps within content for outputting in youtube description. The syntax is same as that of normal timestamps, and in the generated description, the corrections are added after the timestamps.
#1252

```json
"corrections": [
  {
    "time": "02:40",
    "title": "This old example is not using classes in JavaScript. Check the accompanying code for an updated version!"
  }
],
```

```txt
Corrections: 
02:40 This old example is not using classes in JavaScript. Check the accompanying code for an updated version!
```

--- 

## Added support for youtube video urls for `yt-desc`

```
npm run yt-desc https://youtu.be/0V3uYA1hafk
```

----

While working on this, I realised an error in the `yt-desc` script, that it only considers one url for each video, whereas in reality there can be multiple urls for a single video on the site.
For example, it does not recognise the following url
```
https://thecodingtrain.com/tracks/p5-tips-and-tricks/more-p5/pixel-array
``` 
because the canonical url for this video would be 
```
https://thecodingtrain.com/tracks/pixels/more-p5/pixel-array
```
So, in the future, the script have to be reworked to support multiple urls for a video.

### DIFF
--- a/content-testing/schemas.js
+++ b/content-testing/schemas.js
@@ -13,6 +13,7 @@ const {
   youtubePlaylistIdValidator,
   videoNumberValidator,
   timestampsArrayValidator,
+  correctionsArrayValidator,
   urlOrRelativeLinkValidator
 } = require('./validators');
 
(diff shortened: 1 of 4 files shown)


dipamsen:
Btw, I kept the key `"title"` even though it doesn't make much sense to keep things consistent with the normal timestamps. I think it should be fine.

shiffman:
Great, thank you! We could use "text" instead of "title" maybe, but I don't think it's a big deal. Merging! Thank you @dipamsen!

PR #1294, opened 2023-11-09
yt-desc refactor

dipamsen:
Standardised the Video Object, and fixed the issue for supporting non-canonical urls.


shiffman:
Fantastic, thank you @dipamsen!

PR #1305, opened 2023-11-13
fix: updated image and url for challenge flocking contribution 24

vojtatom:
I have updated the image and url for challenge flocking contribution 24 after moving my site to GitHub pages.


shiffman:
Wow, this is incredible, thank you for sharing! Great documentation of your work.

PR #1310, opened 2023-11-16
Fix metadata images

dipamsen:
Fixed guides metadata image to use guide cover image.

old;

new:

Also commented some info on the different image variables for metadata.


dipamsen:
### tests

1. track


2. challenge 



3. track video falls back to track


4. track -> placeholder


5. video -> track -> track placeholder

shiffman:
Oh, so happy for this one! Noticed this issue a while back and forgot about it.

PR #1332, opened 2023-12-15
Add new Coding Train Live! track

kfahn22:


dipamsen:
maybe this doesn't need credits? 😅

dipamsen:
Also, in the future it might be useful to have a "Live" section on the site, or feature this track on the homepage; if there ends up happening multiple project live streams.

kfahn22:
I agree it doesn't need the credits, but it fails the tests w/o them.  Also, I am under the impression that there are a lot of videos on the site that haven't been edited that have the credits.

I am not sure where the best place to put the livestream videos is, but this was an easy way to add the video and allow people to start contributing  their own versions.

dipamsen:
For the credits, I think providing an empty array will pass the tests.

shiffman:
I think this is a great start! Ideally, I would love to annotate and document every single live stream and have a separate section of hte website like challenges, tracks, but I think that's too big of an undertaking and unrealistic to do right now. So I track for "select" livestreams makes a lot of sense! 

I don't feel too strongly about the credits, it's true they don't apply for livestreams so maybe it makes sense to leave a blank array. @kfahn22 let me know when this is ready to merge! I'll review and see if I want to make any small edits too.

kfahn22:
@shiffman I have tried passing a blank array and it is failing the content tests for me.  Other than that, it is ready for review.

fturmel:
> @shiffman I have tried passing a blank array and it is failing the content tests for me. Other than that, it is ready for review.

You can remove the `min(1)` and/or `required()` rules on the credits field of the schema if you want to loosen the validation rule. Make sure to test that the site is able to handle an undefined or empty array value before merging!

https://github.com/CodingTrain/thecodingtrain.com/blob/3f59b50d8c3137a11c1f9083fa49c832145dc263/content-testing/schemas.js#L121-L129

dipamsen:
I have changed the tests to allow for empty credits.

shiffman:
Apologies for the delay, this is wonderful, thank you everyone! I'm going to go ahead and merge and we can always refine more later!

PR #1338, opened 2023-12-22
Add Genuary livestream to CT live track

kfahn22:


dipamsen:
Is the plan for the livestreams to be in reverse chronological order (newest first) in the track? I think that's what makes the most sense probably.

shiffman:
Yes, I think reverse chronological order makes sense! Thank you @kfahn22 for this! Just in time for Genuary this year!

PR #1343, opened 2023-12-26
Adding new logo

runemadsen:
Hi @shiffman! This adds the new logo to the menu bar to replace "THE CODING TRAIN" characters.

If you have time, it might make sense to make a new SVG file that has both the train and the name next to each other. This will make it take up a bit more space. You should be able to just update the SVG file and the CSS should work for this as well. Happy to answer any questions!


shiffman:
Thank you @runemadsen, this is fantastic!! @jasontheillustrator take a look here and let us know if you have any comments before I merge! Do you want to create an SVG with the logo + name as Rune suggests?

https://deploy-preview-1343--codingtrain.netlify.app/

Screen Shot 2023-12-31 at 3 45 19 PM

Issue #1319, opened 2023-11-29
link directly to a "part" of a coding challenge?


shiffman:
I was just linking the TF-IDF challenge from my syllabus and noticed that the URL only goes to part 1 of the challenge

https://thecodingtrain.com/challenges/40-word-counter

I wonder if there's an easy way I could link directly to part 3 (TF-IDF) with something like

`https://thecodingtrain.com/challenges/40-word-counter#part-3`

or

`https://thecodingtrain.com/challenges/40-word-counter#tf-idf`

loic-brtd:
I can try to implement this feature 🙂 I think the easiest way would be to use `#part-3` as it doesn't require us to add (or generate) a slug for each part. Also it's unlikely that the order of the parts will change for existing coding challenges.

PR #1380, opened 2024-01-18
Hash-based challenge parts navigation

loic-brtd:
Choo choo! :steam_locomotive: 

In response to #1319, I'm giving a go at implementing hash-based navigation between challenge parts. This would allow us to reference a specific challenge part by its URL. Example: https://thecodingtrain.com/challenges/40-word-counter#part-3.

A few observations on my current implementation : 
- `/challenges/40-word-counter` takes us to the 1st part of the challenge
- `/challenges/40-word-counter#part-1` takes us to the 1st part of the challenge
- `/challenges/40-word-counter#part-2` takes us to the 2nd part of the challenge
- `/challenges/40-word-counter#part-3` takes us to the 3rd part of the challenge

Whatever hash value is passed to the URL by the user, the URL is not modified (unless we click on another part link or prev/next button of course) : 
- `/challenges/40-word-counter#part-42` takes us to the 3rd (last) part of the challenge (`#part-42` stays in the URL)
- `/challenges/40-word-counter#part-abc` takes us to the 1st part of the challenge (`#part-abc` stays in the URL)

I added this on challenge and track pages containing multi-part challenges.

Previews : 
- https://deploy-preview-1380--codingtrain.netlify.app/challenges/40-word-counter
- https://deploy-preview-1380--codingtrain.netlify.app/tracks/games/72-frogger#part-3

I'm still working on cleaning up some of the code. I welcome any feedback on this feature :grinning:

### DIFF
--- a/src/components/challenges/PartsTimeline.js
+++ b/src/components/challenges/PartsTimeline.js
@@ -4,7 +4,12 @@ import cn from 'classnames';
 import * as css from './PartsTimeline.module.css';
 import { Link } from 'gatsby';
 
-const PartsTimeline = ({ className, parts, currentPartIndex }) => {
+const PartsTimeline = ({
+  className,
+  parts,
+  currentPartIndex,
+  onSelection = () => {}
+}) => {
   return (
     <div className={cn(css.root, className)}>
       <div className={css.partsTimeline}>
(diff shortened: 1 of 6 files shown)


shiffman:
I haven't reviewed the code closely, but I've tested out the implementation in the deploy branch and I think it's working terrifically!  Maybe @fturmel or @dipamsen might have a few minutes to take a look at the code?

loic-brtd:
Thanks @shiffman, I made a finishing touch to the code/documentation and marked this PR as ready for review :smiley:

fturmel:
Fantastic job @loic-brtd, especially noticing the track stops behavior!

I've noticed two things:

* There's a a regression in the mobile view. When you open the panel to navigate between parts, when selecting one the panel doesn't close anymore.
* I personally might prefer getting rid of `#part-1` since it is equivalent to not having an anchor at all

Good stuff!

loic-brtd:
Thanks @fturmel !

> * There's a a regression in the mobile view. When you open the panel to navigate between parts and select one, the panel doesn't close anymore. Same issue in tracks.

Good catch! I'm gonna fix this.

> * I personally might prefer getting rid of `#part-1` since it is equivalent to not having an anchor at all

Why not :) I'd like to have multiple opinions on this

loic-brtd:
@fturmel Btw I just noticed that even on the live website (main branch), the timeline doesn't close on mobile in tracks. Seems like I forgot to handle that case when implementing the feature back then (#743) :upside_down_face: . Anyway, I pushed a fix :smiley:

shiffman:
I think I would agree that the main link to the challenge should just default to part 1 and a separate #part-1 is not needed! However, I'm fine with both working!

loic-brtd:
@fturmel @shiffman I removed `#part-1` in links for the first part of challenges.

I also simplified the logic : if the URL hash value matches `#part-{partNumber}` where `1 <= partNumber <= number of parts` then we're taken to this part, else we're taken to the first part.

fturmel (approved):
Looks great to me!

shiffman:
This is incredible work, thank you for your contribution @loic-brtd! Thrilled to merge this (and thank you for your patience!)

PR #1418, opened 2024-01-29
Fixed retina displays contribution-1706412654.json

tobidot:
New version with 'pixelDensity(1)' fixing the canvas for retina display.


shiffman:
Thanks for your patience, glad this worked @tobidot!

PR #1427, opened 2024-02-01
Jmoptimus showcase submission

JmOptimus:
With this contribution I propose an **extension** of Wolfram Elementary Cellular Automata:

- Different neighbourhoods size (apart of 3).
- Different initial seed (not just ..000010000...).
- Dynamic "world" generation/exploration with user key interactions.


shiffman:
This is fantastic, thank you for sharing @JmOptimus!

PR #1439, opened 2024-02-09
Showcase Language Updates

shiffman:
Some small changes to the submission form as preparation for sharing the passenger showcase on instagram. I'd also like to add a checkbox that allows people to indicate they would not want their project shared outside of the website. Something like:

* [ ] This project is for the Coding Train website only, please do not share on other social media platforms (Instagram).

This would then show up as true or false in the JSON? I'm happy to implement this, but @dipamsen any tips or thoughts about adding this? Would welcome help with implementing it if you are interested!

I could also include opting out of featuring in a live stream, but I don't know if I can commit to that as I browse the website informally during live streams. With instagram it will be curated so we can check.


dipamsen:
Great idea,  though I'm not really sure about opt-in v/s opt-out, as sometimes people may unknowingly check the box, thinking it is some terms and conditions etc. But again, making it opt-in won't be backwards compatible...

shiffman:
Returning back to this, we could use radio buttons that has a default selection, something along the lines of:

🚂 In addition to the website, we are featuring the Coding Train showcase on social media! How would you like to travel?
* [x] The Cafe Car: You have my permission to share this project to other Coding Train social media platforms!
* [ ] By private carriage: I prefer this project to be featured on the Coding Train website only.

dipamsen:
In b4fa227420470f595f571b2a666f50c4d4543ef5 I've added a very hasty attempt at styling the form to show the radio buttons.

- The original design uses `-webkit-appearance: none` to remove default browser styling of these elements to completely restyle them. However no styling is present for radio buttons, as it is not used anywhere on the website. I've just turned on the `-webkit-appearance: radio` for radio buttons, so default browser specific radio buttons are used.
- For the appropriate UX, radio buttons are required to each have a `label` element related to it, so that clicking on the label selects the corresponding option. However, the styling/layout/logic of the `PassengerShowcaseForm` relies on using one `label` for each type of individual input:
  ```html
  <label>
    Label Here <!-- styled with large font -->
    <input />
    <span>Details here</span> <!-- styled as small, muted text -->
  </label>
  ```
  But this system won't quite work for the radio buttons, as we would need the option texts (_"The Cafe Car: ..."_) to be the labels, rather than the name of the input (_"Social Media Permission"_). So, I've had to separate these into different labels, so that the UX is maintained:
  ```html
  <label>Social media permission</label>
  <label>
    <span className={css.radioLabel}>
      <input {...} />
      The Cafe Car: ...
    </span>
  </label>
  <label>
    <span className={css.radioLabel}>
      <input {...} />
      By private carriage: ...
    </span>
  </label>
  ```
  For now, I have used a custom class, to distinguish the other `span`s (muted text), with these, which are the input labels.

dipamsen:
A possible idea could be to show the `Twitter` and `Instagram` inputs, only if _"The Cafe Car"_ is selected for the permission.

fturmel:
Should this new property be added to the content tests schema?

https://github.com/CodingTrain/thecodingtrain.com/blob/3f99835f97b859fa852de8553fe404c2998da59c/content-testing/schemas.js#L148-L158

Are previous showcase submissions opted-in automatically, opted-out, left as undefined? (and should they be backfilled so this new value can be required both in the submission form and pull requests?).

fturmel:
> Opted-in automatically, and absense of the field implies opt-in (no back-filling required)

If we don't back fill, it means this field needs to be optional in the content tests. Showcases contributed as GitHub PRs might not realize that omitting/forgetting this new field in the JSON file implies an opt-in. Something to consider...

shiffman:
> If we don't back fill, it means this field needs to be optional in the content tests. Showcases contributed as GitHub PRs might not realize that omitting/forgetting this new field in the JSON file implies an opt-in. Something to consider...

Thanks for this feedback @fturmel, I'm not sure the best way to handle this. On the one hand, I could make the argument that this field is unnecessary and be more clear that projects submitted to the website may be featured during live streams and other other social platforms (w/ credit). I think that someone who wouldn't want their work seen wouldn't want it on the website in the first place. But I think having the option to check is helpful and important?

I don't know whether backfilling would make sense, it feels weird to add "true" for someone who didn't themselves, I'm probably overthinking this, but we could backfill with `null` or something lol? So, it's basically "yes", "no", or "older submission, grandfathered in"? I'm going to go ahead and merge this as we're starting the new passenger showcase sharing initiative tomorrow! We can discuss this in a separate issue / branch?

Issue #131, opened 2022-05-12
Finalize Acknowledgements


shiffman:
I am working on the names and roles here:
https://docs.google.com/document/d/1Yw79zplmjslIGCsikg-MWD_ZARcPSeP5-kH524WKEDc/edit?usp=sharing

@fdoflorenzano mentioned during our meeting this could use some design cleanup as well.

fdoflorenzano:
Nice! Thanks for getting that going!

We're thinking of removing a group that only contained DSI, and maybe add a mention in the footer. 

You can leave DSI just as a friend of The Coding Train in this section, and we'll add it also in another section.

shiffman:
Started #240 to address this. Also, there are now several generous amazing volunteer contributors to the site from the Coding Train community. Too many for me to mention but @kfahn22 @skaser85 and @dipamsen have been particularly active recently. What do you all think about adding https://allcontributors.org/ to this repo for additional community acknowledgement and credit?

Open to any and all thoughts here!

shiffman:
Also @KobeLiesenborgs would you be able to add YT and Discord moderation team members or provide me with a list?

shiffman:
I've added `allcontributors`, so the only outstanding item here is the YT and Discord moderators, @KobeLiesenborgs can you help with this last step?

shiffman:
It took me so long to finalize that now this page needs a whole set of updates! @Muntasir2001 could you help me with this as it relates to Discord and moderators? I'll start a new branch.

Muntasir2001:
@shiffman just to clarify  is this related to this page?

![image](https://github.com/CodingTrain/thecodingtrain.com/assets/52258261/0fc07e94-e9b4-4561-9a43-c615a096b727)

(or I am a bit confused here about what's going on 😅 )

shiffman:
Correct! The elements for you to look at is how you would like to be listed under "Team" and helping list the current active moderators under "Moderators". You can see the changes I've made so far here: https://github.com/CodingTrain/thecodingtrain.com/pull/1514/files

You can reply here (or we can discuss over DM in Discord) or make changes to the pull request branch directly!

Muntasir2001:
Makes sense

I will have a look and let you know :)
Probably will make change to the PR directly as it saves time

Muntasir2001:
[...]

PR #1514, opened 2024-03-19
Acknowledgements Updates

shiffman:
See #131


shiffman:
Thank you @Muntasir2001!

PR #1516, opened 2024-03-19
Remove redundant video titles from descriptions

### LINKED ISSUE #1280: simplify YT links in description generator
I noticed how that YT renders the YT icon + video/playlist title in the description itself. So having the additional text is likely redundant:

Screen Shot 2023-11-03 at 6 35 14 PM
Screen Shot 2023-11-03 at 6 35 09 PM

dipamsen:
Fixes #1280 

Old:

image

New:

image

Kept `CC###` for the related challenges, because sometimes the challenge titles may not have necessary context (eg. _"What was Coding like 40 years ago?"_)

----

Do these also need to be changed?

image

### DIFF
--- a/content/videos/challenges/171-wave-function-collapse/index.json
+++ b/content/videos/challenges/171-wave-function-collapse/index.json
@@ -47,7 +47,7 @@
       "description": "GitHub repo for expanded and corrected WFC PCB board example.",
       "image": "wfc.jpg",
       "urls": {
-        "github": "https://github.com/CodingTrain/Wave-Function-Collapse"
+        "other": "https://github.com/CodingTrain/Wave-Function-Collapse"
       }
     }
   ],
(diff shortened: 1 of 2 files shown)


shiffman:
@dipamsen yes, it would probably make sense to do the playlists too! I don't think we need the `CC##`, the point is to just reference and link to the video I don't think anyone is obsessively tracking the challenge numbers in the description?

shiffman:
I think for previous/next/all we need to keep that text but can remove the word "video" to be more concise. For other parts we don't need the title just the video link. Thank you!!!

dipamsen:
I think I've done the necessary changes, here's a comparison (handmade) based on the changes for the WFC description:

Before:

image

After:

image

(I am now noticing the missing `:` on the link title, which i'm not quite sure how it happened)

dipamsen:
(fixed the colon)

dipamsen:
cc @shiffman

shiffman:
Thanks @dipamsen! Merging, we can make further adjustments as needed! I'm not going back and updating old descriptions but will try this out for the next challenge!

PR #1532, opened 2024-03-25
Improved responsiveness of Showcase form & removed empty ruleset.

Zack-Dx:
Hello, I noticed that the submission form select and fields were not responsive for a device with a maximum width of 400px.

Before:

Before

After I fixed it, ensuring the rest of the responsive elements are in place:

After

* Also eliminated a single unneeded ruleset.


shiffman:
Thank you for this, wow! I will test this soon and hopefully merge! @dipamsen or @fturmel feel free to take a look too if you have a minute!

fturmel:
I like it! Could we try using 100% instead of arbitrary pixel values, and apply them to the --small (600px) breakpoint instead? That way the fields will fill up the width more consistently and we'll get rid of the remaining broken state between 400px and ~525px browser width.

As for removing the empty root css class in the tracks Header component, it wasn't bothering me since most components are structured with a root class and roughly 5 of them are empty.

If we do want to remove the empty class properly though, the reference to it in the component should also be deleted:

https://github.com/CodingTrain/thecodingtrain.com/blob/fcddb57a2b684a908e5231c114c4dfe9df7cee32/src/components/tracks/Header.js#L14

shiffman:
Hi @Zack-Dx did you mean to close this? I would like to incorporate it! Apologies for not following up about it sooner.

Zack-Dx:
> Hi @Zack-Dx did you mean to close this? I would like to incorporate it! Apologies for not following up about it sooner.

No issues Shiffman, I'll create a new pr fixing this I'm just busy these days a bit due to my exams, I'll fix by upcoming weekend. Thanks love you.

Issue #1550, opened 2024-04-01
Prevent Duplicate Showcase Submissions


dipamsen:
Maybe we could add something like "will be reviewed within [timeframe]" or "Please do not submit twice" to prevent duplicate submissions.

Another idea could be to link to /pulls so that the user can be convinced that the pr has been created

fturmel:
@dipamsen There _has_ been a lot of dupes lately.

I agree the messaging could be improved (it's currently "Thank you for submitting to the Passenger Showcase! Please refresh the page in order to upload another submission."), but I think it could be an implementation issue on our end.

Look at #1588 and #1589 for example, the `submittedOn` server timestamps are about 1 second apart. That's too short for a user to fill the form again and re-submit.

At a glance, one possible reason would be because there's an onClick handle on the submit button AND an onSubmit on the form. The submit button should be of type "submit" and not have a click handler. Maybe in some cases both the click and form submit fire at the same time.

https://github.com/CodingTrain/thecodingtrain.com/blob/eb0535a049e9968d5a3fc53f5ecb79d48f673b52/src/components/PassengerShowcaseForm.js#L225

https://github.com/CodingTrain/thecodingtrain.com/blob/eb0535a049e9968d5a3fc53f5ecb79d48f673b52/src/components/PassengerShowcaseForm.js#L428-L435

fturmel:
The `submitted` state is also only updated once the response comes back from the API call. We might need a `waiting` interim state to disable the form/button and some visual indicator that the request is being processed.

shiffman:
Also, it might be nice to add something more unique (beyond just a reference to the video) to the pull request subject, this will make it easier for me to recognize dupes?

fturmel:
Cool, I can take a stab at this. I have some additional ideas on how to improve the form experience as well. We can fine-tune and discuss the specifics in the pull request.

kfahn22:
I am also wondering if there is an issue with pull requests disappearing.  I didn't notice until just now, but I submitted two for marbling and there is no record of the second one.

fturmel:
> I am also wondering if there is an issue with pull requests disappearing. I didn't notice until just now, but I submitted two for marbling and there is no record of the second one.

Can you share more details? Was it through the Coding Train web form? Were they submitted the same day? Does it go through if you try again with the same image/info?

I checked the Netlify functions log history and I could see the one from May 2nd that got merged but not the other. The logs only go back a week though.

[...]

PR #1605, opened 2024-05-19
Add waiting state to showcase form

dipamsen:
See #1550 

- Adds "waiting" state to showcase form to prevent duplicate submission from double clicks
- Style `disabled` button properly 
 
cc @fturmel

### DIFF
--- a/src/components/Button.module.css
+++ b/src/components/Button.module.css
@@ -137,11 +137,16 @@ button.rainbow.cyan {
   transition: background-position 1s ease-out;
 }
 
-.rainbow.red:hover,
-.rainbow.orange:hover,
-.rainbow.purple:hover,
-.rainbow.pink:hover,
-.rainbow.cyan:hover {
+.rainbow.red:hover:not([disabled]),
+.rainbow.orange:hover:not([disabled]),
+.rainbow.purple:hover:not([disabled]),
+.rainbow.pink:hover:not([disabled]),
+.rainbow.cyan:hover:not([disabled]) {
   background-position: left bottom;
   transition: background-position 4s ease-out;
+}
+
+button[disabled] {
+  opacity: 0.5;
+  cursor: not-allowed;
 }
--- a/src/components/PassengerShowcaseForm.js
+++ b/src/components/PassengerShowcaseForm.js
@@ -132,7 +132,7 @@ const PassengerShowcaseForm = () => {
     video: location.state?.video ?? ''
   });
   const [error, setError] = useState(null);
-  const [submitted, setSubmitted] = useState(false);
+  const [submitted, setSubmitted] = useState('not-submitted'); // not-submitted, waiting, submitted
 
   const data = useVideosWithShowcase();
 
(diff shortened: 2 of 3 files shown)


src/components/Button.module.css
fturmel, line 1: Can we use the pseudo class `:disabled` instead of the attribute `disabled`? We might not need all the color variants specificity here either.

```css
.rainbow:hover:not(:disabled) {
  background-position: left bottom;
  transition: background-position 4s ease-out;
}

button:disabled {
  opacity: 0.5;
  cursor: not-allowed;
}
```

And well done on the negation here, I was wondering why you didn't use `:enabled` initially but this button component is sneaky and can end up being a `<button>` or a `<Link>` / `<a>` tag based on the props passed to it.

src/components/PassengerShowcaseForm.module.css
```diff
@@ -64,6 +64,6 @@
   }
 }
 
-.submitBtn:hover {
+.submitBtn:hover:not([disabled]) {
```
fturmel, line 67: Here we could actually use `:enabled` since it is always a button

shiffman:
Looks great, thank you @dipamsen! Tagging in @fturmel in case you have a chance to review!

src/components/PassengerShowcaseForm.js
```diff
@@ -429,7 +435,7 @@ const PassengerShowcaseForm = () => {
           className={css.submitBtn}
           onClick={onSubmit}
           variant="purple"
-          disabled={submitted}
+          disabled={submitted !== 'not-submitted'}
```
fturmel, line 438: Nit-pick but I find it hard to parse cognitively with the double (triple?) negative.

Maybe renaming the variable to `submissionState` and finding an alternate name for `not-submitted` state that doesn't contain a negative would help? What about `undefined`, would that read better?

```ts
submissionState === 'submitted'

disabled = !!submissionState
```

fturmel:
Looks great, I agree!

Sorry for not shipping a fix myself after self-assigning the issue. I've been taking notes on additional improvements to the submission process and did not get a chance to code or share anything yet. More soon-ish!

src/components/Button.module.css
dipamsen: whoops, didn't realise i was inconsistently using `:disabled` and `[disabled]` in different places

fturmel: Thought about this a bit more, and maybe the `:not(:disabled)` bit is too "clever" without a comment? Maybe something like this would be clearer/more self-documenting in the long run.

```css
a.rainbow, button.rainbow:enabled {
  &:hover {
    background-position: left bottom;
    transition: background-position 4s ease-out;
  }
}
```

shiffman:
Thank you @dipamsen, excited to merge this!

dipamsen:
Oh whoops looks like the review comments were pending to be addressed...

shiffman:
Oh, sorry about that! Would you like me to revert or shall we just start a new branch? Apologies for this mistake!

fturmel:
No worries, and no need to revert since the comments were more about minor code quality improvements. I can address them in an upcoming PR.

PR #1554, opened 2024-04-14
Added Dark Mode Feature to thecodingtrain Website 

Yashasewi:
# Dark Mode Implementation

This pull request introduces the implementation of the dark mode feature for the thecodingtrain.com website, as discussed in #1504.

## Changes

- Implemented dark mode using CSS variables and theming classes
- Created a dedicated `theme.js` component to change CSS variables for light and dark modes
- Updated existing styles to reference the appropriate CSS variables for theming
- Addressed instances of hard-coded colors and updated them to use CSS variables

### CSS Variables and Theming

The `theme.js` file defines two sets of CSS variables: one for the light mode and another for the dark mode. These variables control various colors and styles throughout the website, including text, backgrounds, and UI elements.

```Javascript
/* Light Mode Variables */
const lightTheme = [
  '--text-color: #000000',
  '--gray-dark: #3c3c3c',
  '--gray-mid: #666666',
  ......
];

/* Dark Mode */
const darkTheme = [
  '--text-color: #f0f0f0',
  '--background-color: #1e1d20',
  '--gray-dark: #1f1f1f',
  '--gray-mid: #b3b3b3',
  ......
];
```

The `theme.js` component manages the application of the appropriate CSS variables based on the user's preference (light or dark mode). It also persists the user's preference in localStorage for consistency across sessions.

### Existing Styles Updates

Existing CSS styles have been updated to reference the appropriate CSS variables, ensuring that colors and visual elements adapt seamlessly to the selected theme.

## Potential Issues

-  While updating the styles to use CSS variables, there might be some instances where colors were hard-coded or missed. These instances could potentially lead to inconsistencies or unintended visual effects in the dark mode. Further testing and review from the team would be beneficial to identify and address such cases.
-  Additionally, the icons in the footer cannot be changed by simply adjusting the color, so we might need to use new icons or find an alternative solution to ensure they are visible and consistent with the dark mode.
- Regarding the gradient present on the home page's homescene, I couldn't find a suitable alternative that works well with the dark mode. I have left this part for the design team to decide on a new gradient or alternative design element that looks good in the dark mode.

## Next Steps

- Address any feedback or suggestions from the review process
- Conduct accessibility audits to ensure compliance with WCAG guidelines
- Identify and address any remaining instances of hard-coded colors or inconsistencies
- Please review the changes and provide your feedback. I'm happy to address any concerns or make further adjustments as needed.

Resolves #1504

@shiffman @runemadsen @jasontheillustrator

PR #1607, opened 2024-05-20
Fix table overflow issue on Chromium browsers

Yashasewi:
## Description

This PR fixes an issue where the table on the "Maintaining the Website Guide" page overflows due to the YouTube URL not wrapping on Chromium-based browsers. The issue was mentioned in [this comment](https://github.com/CodingTrain/thecodingtrain.com/pull/1554#issuecomment-2097954963).

## Steps to Reproduce

1. Open the "Maintaining the Website Guide" page on a Chromium-based browser (e.g., Google Chrome, DuckDuckGo, Brave) using a desktop or mobile device. [here](https://thecodingtrain.com/guides/maintain-website-guide)
2. Scroll down to the table section.

## Expected Behavior

The table should display correctly, with the YouTube URLs wrapping to the next line if they exceed the table cell width.

## Actual Behavior

On Chromium-based browsers, the YouTube URLs do not wrap, causing the table to overflow horizontally.

## Screenshots

| Before | After |
|:------:|:-----:|
| Before | After |

## Testing

- [x] Tested on Google Chrome (Android)
- [x] Tested on DuckDuckGo (Android)
- [x] Tested on Brave (Android)
- [x] Tested on Mozilla Firefox (Android) (has no effect)
- [ ] Tested on Safari (iOS) [If you have access to an iPhone, please confirm and check this box]

### DIFF
--- a/src/templates/guide.module.css
+++ b/src/templates/guide.module.css
@@ -189,6 +189,10 @@
   padding: 0 var(--spacing-xsmall);
 }
 
+.td:not(:first-child) {
+  word-break: break-word;
+}
+
 .tr:last-child .td {
   border-bottom: none;
 }
@@ -214,12 +218,6 @@
   .videoContainer {
     padding-bottom: var(--baseline-box-7x);
   }
-  .table {
-    & table {
-      table-layout: fixed;
-      text-wrap: wrap;
-    }
-  }
 }
 
 @media (--xsmall) {
@@ -228,11 +226,5 @@
   }
   .itemSpacer {
     display: none;
-  }
-  .table {
-    & table {
-      table-layout: fixed;
-      text-wrap: wrap;
-    }
   }
 }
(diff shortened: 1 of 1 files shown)


dipamsen:
Are the before after images swapped? 😄

Yashasewi:
lol @dipamsen ....

Yashasewi:
> Are the before after images swapped? 😄

Now it's looking good … 😂

fturmel:
This is not working properly for all screen sizes:



Also `table-layout:fixed` makes all columns equal width, and that's not the optimal use of horizontal space with the type of content that ends up in these tables.

I would propose keeping the table auto layout and applying word breaks in td elements that are not the first column, since those tend to be single words. I think this is all we would need:

```css
.td:not(:first-child) {
  word-break: break-word;
}
```

Yashasewi:
@fturmel Thank you for the proposed solution. I agree that using  table-layout:fixed is not the best approach here . Updated with new solution .

fturmel (approved):
Looks good to me, thanks!

shiffman:
Amazing, thank you!!

PR #1621, opened 2024-06-14
Auto update YT Descriptions using the YouTube API

dipamsen:
For now, this script only allows updating one video at a time, because of safety reasons, as well as the YouTube `videos.update` API has a high quota cost.

Obviously, this is untested code. (I have only tested the `updateDescription` function with a video on my channel.)


shiffman:
Hi @dipamsen, super excited about this! How about we set up a "dummy" channel to test this out on before deploying it with the CT channel itself? We could maybe hardcode a single video ID and use that over and over again? Let me know, I can work on that!

dipamsen:
The current script is coded such that it only updates a single description at a time (to minimise damage), I think it can be directly tested on the actual channel on a obscure/non popular video, and in the worst case scenario where the description is lost, it can be manually restored via the generated description (yt-desc).

This way the entire workflow can be tested which involves getting the yt I'd of the video etc

shiffman:
Apologies for the delay @dipamsen I hope to try this out soon! Thanks for your patience!

fturmel:
@dipamsen Great job on this.

I had a thought earlier and wanted to share, feel free to dismiss!

The Coding Train channel has 1266 videos total (more than we have integrated in the site for sure). Since `video.list()` is cheap (1 quota unit) and you should be able to give it up to 50 ids per call, it would only cost up to 26 units to fetch the description of every single video.

From there, you could only offer to update the videos that have a different description that the one expected. Maybe you could even show a little diff between the two strings in the terminal before going through with the update?

PR #1631, opened 2024-06-23
Minor URL and copy updates

fturmel:
A few small things I noticed.

* The internal Discord links are now relative
* The 2023 workflow link in the FAQ was wrong
* Misc text and URL updates


shiffman:
Thank you for catching all these @fturmel and thank you for your patience!

PR #1632, opened 2024-06-24
Update Node.js to 20 LTS

fturmel:
With 22 recently released, 20 is now the new active LTS. It's also stable on AWS lambda so the timing should be good to migrate. We should be doing this mostly to keep up with the ecosystem and get some free perf improvements.

Just before merging this PR, the environment variable `AWS_LAMBDA_JS_RUNTIME` on Netlify should also be set to `nodejs20.x`

https://nodejs.org/en/about/previous-releases
https://docs.aws.amazon.com/lambda/latest/dg/lambda-runtimes.html


shiffman:
I've updated the environment variable, merging, thank you @fturmel!

fturmel:
@shiffman Last time we upgraded Node was in late 2022, and I just realized that there has been a [product update](https://answers.netlify.com/t/changes-to-the-default-node-js-version-for-netlify-functions/90980) since then that makes the use of the environment variable redundant. Netlify Functions will inherit the version defined in the `.nvmrc` file as long as it is supported.

shiffman:
Ah, thank you @fturmel, good to know! Would you advise that I remove the environment variable or leave it for now?

fturmel:
> Ah, thank you @fturmel, good to know! Would you advise that I remove the environment variable or leave it for now?

We can remove it, the simpler the configuration the better!

PR #1646, opened 2024-07-05
Update non-challenge video thumbnails + automation script

### LINKED ISSUE #1645: Nature of Code track video preview images
This is a small thing but I'm adding links from natureofcode.com to here (see: https://github.com/nature-of-code/noc-book-2/pull/985) and noticed that the NOC videos are just using the default site image and it would be nice to update them at some point!

image

fturmel:
closes #1645


shiffman:
Ah, this script is so great, thank you @fturmel! Should I merge or are we still discussing / thinking as per #1645?

fturmel:
Your call, we've outlined the options in the issue discussion. I don't mind if we close this PR in favor of using the track thumbnail for all non-challenge videos if that's a better fit.

Having thumbnails for each video is not a bad thing though. Given that the individual videos are substantial, the intent when sharing a link is most likely to reference the video and not necessarily the parent track.

fturmel:
Closing in favor of #1648

PR #1648, opened 2024-07-06
Delete misplaced placeholder images for track videos

dipamsen:
Alternative to #1646 .

See https://github.com/CodingTrain/thecodingtrain.com/issues/1645#issuecomment-2211483499

Fixed preview image for "Vector Math" - falls back to NoC track thumbnail

image


shiffman:
Merging as per discussion in #1645

PR #1652, opened 2024-07-11
Improve sort order of `Select` component filtered results

### LINKED ISSUE #1264: bug/feat: exact match should be first result in search dropdown
image

Here, preferably "data" should be listed as the first search result, or it leads to a weird interaction where typing "data" and <kbd>Enter</kbd> opens up the search results for "COCO dataset".

fturmel:
closes #1264

Re-implements the `react-select` default filtering logic to be able to apply better ranked sorting on results. This applies to all instances of the `Select` component (tracks, challenges and showcase pages).

Sort priority:
* Exact match
* Exact word match + how close to start of string the match is
* Partial match + how close to start of string the match is

| Before    | This PR |
| - | - |
| CleanShot 2024-07-11 at 12 16 01@2x  | CleanShot 2024-07-11 at 12 16 37@2x |

### DIFF
--- a/src/components/Select.js
+++ b/src/components/Select.js
@@ -1,10 +1,8 @@
-import React, { useMemo } from 'react';
+import React, { useMemo, useState } from 'react';
 import ReactSelect from 'react-select';
 import cn from 'classnames';
 
 import * as css from './Select.module.css';
-
-const toOption = (value) => ({ value, label: value });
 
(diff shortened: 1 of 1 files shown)


shiffman:
Tested and working beautifully! Reminds me that I should periodically do a round of cleanup and alignment of the topics/tags to be more consistent.

Issue #46, opened 2022-01-19
Track Overview Scrolling


NwE0kmCE:
On tracks that have enough videos to fill the sidebar, scrolling down and clicking on a video scrolls to the top of the video list:
![Scrolling](https://user-images.githubusercontent.com/28796123/150204139-b11a8dbb-80c4-4e3a-8e66-da77db91802b.gif)
.

PR #1653, opened 2024-07-11
Remove passive event "polyfill" and unused `randomElement()` utility

fturmel:
We can remove this passive event listener polyfill thing - I don't think it ever was needed as it was introduced in 2022 (#82) but only exists because of old/dead Edge browser (pre Chromium I assume) from before 2017 and IE 11.
https://caniuse.com/passive-event-listener

I corrected the listeners usage too. The `capture` option is `false` by default, and the `passive` option is not required when removing the listener (it's not even a valid option to begin with).
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#passive
https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#parameters

I also noticed the `randomElement` utility was never used and can be safely removed from the codebase.

### DIFF
--- a/src/hooks/index.js
+++ b/src/hooks/index.js
@@ -1,6 +1,5 @@
 import React, { useMemo, useRef, useEffect, useState } from 'react';
 import { useLocation } from '@reach/router';
-import { passiveEventArg } from '../utils';
 
 /**
  * Takes an array of images nodes and makes a hashed object based on their names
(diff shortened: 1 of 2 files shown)


shiffman:
Hi @fturmel are there any pages I should load and test to confirm that these changes are working as intended! All looks great from my end but just want to double check before merging!

fturmel:
You could look at the only feature connected to the passive event, which is persisting the scroll position of the content panel on the right side of the video component. See #82 and #46 for historical context.

This track for example: https://deploy-preview-1653--codingtrain.netlify.app/tracks/data-and-apis-in-javascript

https://github.com/user-attachments/assets/350a599e-2128-453d-86b3-7cd8c8f3ec48

shiffman:
Great, thanks for this update and improvement!!

PR #1655, opened 2024-07-12
Add 184-elastic collisions challenge

kfahn22:
This is still missing:

- index image
- description
- videoID
- link to Dipam's paper
- link to other paper mentioned

Date is currently set to today--will need to be updated.

PR #1656, opened 2024-07-12
Improve content test for `videoId` field

fturmel:
#1655 made me realize that there was an error in the way the video schema was defined, and it was allowing empty `videoId` to pass the tests.

The check was also updated so that the top-level `videoId` must be empty or undefined instead of optional when a video is multi-part. It should help avoid confusion and information repetition.

### DIFF
--- a/content-testing/schemas.js
+++ b/content-testing/schemas.js
@@ -21,9 +21,15 @@ const baseVideosSchema = strictObject({
   title: string().required(),
   description: string().required(),
 
-  videoId: youtubeIdValidator.when(['parts'], (parts, schema) => {
+  videoId: youtubeIdValidator.when('parts', ([parts], schema) => {
     // videoId required if the `parts` array is missing or empty
-    return parts && parts.length > 0 ? schema : schema.required();
+    return parts && parts.length > 0
+      ? schema.test(
+          'videoId and parts conflict',
+          'Top-level "videoId" should not be defined if "parts" exist',
+          (value) => !value
+        )
+      : schema.required();
   }),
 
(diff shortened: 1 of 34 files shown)


shiffman:
Hi @fturmel thanks again! Am I confused or is there a mistake in this PR where a bunch of videoIds are removed? Perhaps that is leftover from testing the tests themselves?

shiffman:
Hi @fturmel thanks again! Am I confused or is there a mistake in this PR where a bunch of videoIds are removed? Perhaps that is leftover from testing the tests themselves?

fturmel:
No it's intended, sorry I wasn't clear. There's a bug fix and an improvement in there:

* The existing schema validation was supposed to make sure that `videoId` was required, but the "parts" argument was not destructured from the argument properly. It's basically a `Yup` library API usage mistake.

* The top-level videoIds removed are all for multi-part videos and were effectively useless. I added a check in the content tests to require them to be undefined or empty when a "parts" array exist. Currently they were optional but you could still put a YouTube ID in that field even if it did nothing.

shiffman:
Ah, I see! Thank you for the clarification. Would it be more clear to remove the field entirely rather than have it as a blank entry?

fturmel:
> Ah, I see! Thank you for the clarification. Would it be more clear to remove the field entirely rather than have it as a blank entry?

Sure, we can make the check more strict instead of being falsy (which allows undefined, null or empty string). Update incoming.

fturmel:
Done, the `videoId` field cannot be present anymore if the video is multi-part.

shiffman:
Fantastic!

PR #1624, opened 2024-06-16
Voronoi game of life submission

okg21:
Voronoi game of life showcase submission

PR #1665, opened 2024-07-15
Fix contribution url

fturmel:
Closes #1624

Updated the URL to link to demo, per author request.


shiffman:
Thank you @fturmel!

PR #1690, opened 2024-08-18
Update homepage 8-18-24

kfahn22:


shiffman:
Thank you as always @kfahn22!!

PR #1710, opened 2024-09-01
Updating Amazon Referral Links

shiffman:
I have a new Amazon affiliate account that I set up for the Nature of Code book so I'm updating lots of old links.


shiffman:
Holding off on this b/c I'm having issues with Amazon, my account keeps getting terminated I finally had it re-instated yesterday and then all of a sudden it was terminated again this morning super frustrating!

shiffman:
This should be all set now.

Issue #1730, opened 2024-10-04
Redo ml5.js track and archive old videos?


shiffman:
A new "image classification" video has been recorded. Here is a to do list, would love help with any of this!

* [x] create page for new image classification video
* [x] create "archive" track (put at the end of the tracks page) for out of date videos that have been replaced, starting with these two: [ml5 image classification 1](https://thecodingtrain.com/tracks/ml5js-beginners-guide/ml5/1-classification/1-image-classification) and [ml5 image classification 2](https://thecodingtrain.com/tracks/ml5js-beginners-guide/ml5/1-classification/2-webcam-classification)
* [ ] Update video descriptions for archived videos to point to new ones
* [ ] update all the ml5 folders to take out the numbering, update all the video description on YouTube with new links
* [ ] create redirects for the old urls

We could skip removing the numbers since they don't actually show up on the website, only in the URLs and maybe it's best to just keep the URLs?

PR #1731, opened 2024-10-04
add new ml5 imageclassifier video

dipamsen:
#1730


kfahn22:
Still needs description

kfahn22:
The date and video ID need to be updated before merging.

shiffman:
The video is still unlisted and I probably won't officially release it until maybe even Tuesday, but I'm going to go ahead and merge this so that we can test out this new "archiving" process and I can continue to update other video descriptions on YT that will need to change due to new URLs, etc.

PR #1746, opened 2024-10-14
Minor `about` page image tweaks

fturmel:
Dimensions should be more appropriate for usage now. Note that they were saved as quality 100 since Gatsby will compress them further during build anyways.


shiffman:
Thanks for this one too!

PR #1748, opened 2024-10-16
Add new ml5 Pose Detection video

kfahn22:
Still Needed:  index image

I tentatively moved the showcases from archived video to new video.  I also put the archived ml5 videos in a folder within archives.  I didn't add the CT sketches with "test" in the title, since I was assuming they were for testing purposes (but am not am not positive about that).

PR #1749, opened 2024-10-16
Additional ml5.js track work

shiffman:
This is building on top of #1748 cc @kfahn22


shiffman:
Well, I made a mess of this but i'm going to publish the video so here we go!!!

PR #1755, opened 2024-10-28
Add handPose video

kfahn22:
All the code images and the index image are still needed.  It is currently added as a separate folder in ml5 without any number.  The video still needs to be added to track.  Currently failing because I set the date for tomorrow.


shiffman:
Thank you so much @kfahn22, the starting skeleton is so incredibly helpful!! I've added code examples and additional references, links, etc.

PR #1745, opened 2024-10-14
SVG optimizations

fturmel:
I noticed there was an `svgo` script buried in the code base. I rewrote it so we could optimize all SVGs instead of just the ones in `src/images/characters`. Also the newer version of `svgo` should produce better outputs, the previously minified images were generated with v2.8 and without the multi-pass flag.

We're generating so many static pages and the majority of the SVGs are inlined, so even small improvements here could have some significance.

<details>

<summary>`node node-scripts/minify-svgs.mjs` output:</summary>

```
src/images/tracks.svg
  🟢 -54.23%
     -1.87 kB
static/images/train.svg
  🟢 -4.36%
     -2.45 kB
static/images/nebula-button.svg
  🟢 -14.94%
     -0.97 kB
src/images/sun.svg
  🟢 -43.85%
     -1.24 kB
src/images/playbutton.svg
  🟢 -32.49%
     -0.14 kB
src/images/train.svg
  🟢 -4.36%
     -2.45 kB
src/images/processing-icon.svg
  🟢 -26.79%
     -0.10 kB
src/images/open.svg
  🟢 -20.78%
     -0.07 kB
src/images/p5js-icon.svg
  🟢 -59.33%
     -3.25 kB
src/images/play.svg
  🟢 -22.19%
     -0.08 kB
src/images/playbutton-filled.svg
  🟢 -20.37%
     -0.03 kB
src/images/node-icon.svg
  🟢 -0.68%
     -0.01 kB
src/images/clock.svg
  🟢 -49.53%
     -1.85 kB
src/images/cloud.svg
  🟢 -41.57%
     -0.20 kB
src/images/train_track.svg
  🟢 -67.77%
     -1.20 kB
src/images/new-logo.svg
  🟢 -16.33%
     -0.47 kB
src/images/characters/Zero_6.svg
  🟢 -14.01%
     -0.24 kB
src/images/characters/Zero_3.svg
  🟢 -7.83%
     -0.15 kB
src/images/logo.svg
  🟢 -37.89%
     -4.27 kB
src/images/characters/Zero_5.svg
  🟢 -4.15%
     -0.09 kB
src/images/characters/Zero_2.svg
  🟢 -7.47%
     -0.19 kB
src/images/characters/Zero_1.svg
  🟢 -3.87%
     -0.06 kB
src/images/characters/WheelstheMouse_3.svg
  🟡 0.27%
     0.01 kB
src/images/characters/Zero_4.svg
  🟢 -9.50%
     -0.18 kB
[...]

PR #1760, opened 2024-11-06
Keep `viewBox` attribute on optimized SVGs

### LINKED ISSUE #1759: "The Coding Train" Footer Logo Being Cut Off.
On all of the pages, the text that says "The Coding Train" in the footer is cut off at the word "Train". Instead of the width of the svg being 326px, it is 250px.

Screenshot 2024-11-07 at 10-17-59 The Coding Train

Screenshot 2024-11-07 101722

fturmel:
Closes #1759

This was introduced with #1745, `svgo` defaults strip-out the `viewBox` attribute, which is functional and not a safe optimization to blindly apply.

For reference: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/viewBox

> When an SVG contains a viewBox attribute (often in combination with a preserveAspectRatio attribute), a transform stretches or resizes the SVG viewport to fit a particular container element.


coderpro1234-2:
Preview fixes issue #1759

shiffman (commented):
Thank you @fturmel apologies it took me so long to review this. The fix looks great to me, really appreciate you taking the time!

PR #1762, opened 2024-11-08
Zishan showcase submission

1337-pr05:
this is my work on the space invaders game a late 19s famous game!
my game involves great graphical items + sound effects are added to create a real-life feel on done on using th p5.js framework!

PR #1767, opened 2024-11-15
Zishan remove image

shiffman:


fturmel:
@shiffman looks like there were no changes in this PR when diff'd against `main`, but it was intended to delete a duplicate image per https://github.com/CodingTrain/thecodingtrain.com/pull/1762#issuecomment-2477782845.

I added a commit to properly remove the image.

shiffman:
weird! I have this commit which should in theory have removed the image! 82e6d34b8c7b6a36371224cf2da79cecccad55fb

Thank you so much for fixing this!!

PR #1770, opened 2024-11-16
Run `typos` with a GitHub action on PRs

fturmel:
False positives are infrequent enough that I think we could give this a shot. Let me know what you think.

https://github.com/crate-ci/typos


shiffman:
This is great, totally worth doing! The main thing that comes up with something like this is that the YT descriptions won't be fixed and I still have to manually do them. But that's mostly on me b/c @dipamsen did excellent work in #1621 and I really need to figure out a process for running it.

fturmel:
The hope is that with `typos` running against PRs automatically, we'll be able to catch these mistakes before the YouTube description gets set in the first place.

PR #1799, opened 2025-01-20
Fix footer React hydration error

fturmel:
Happy new year! Last deployment was in 2024, and the mismatched server/client current year has surfaced a little React hydration error.

CleanShot 2025-01-20 at 09 11 48@2x


shiffman:
Thank you for catching this!!

PR #1802, opened 2025-01-24
Add WFC-overlapping video

kfahn22:
Notes:  there is only one code example image, not sure if more are needed.
Still needed: video description, video id, index image, update date, update Nebula link


shiffman:
The video is now set to members only on YT so time to merge! I welcome additional contributions and updates to the challenge page.

PR #1828, opened 2025-02-23
fix url to derivation of collision formula

dipamsen:


shiffman:
Thanks for the update @dipamsen!

PR #1849, opened 2025-05-04
Showcase Submission for code/1-intro/3-shapes-drawing

ch00ooch00oo:
Thank you nhi for your contribution! A member of the Coding Train team will review it shortly.

* [My family portrait](http://buinhi.com)
* [nhi](http://buinhi.com)

preview image


shiffman:
This is so creative! It's my preference that the showcase project links to the p5.js sketch and not a general portfolio, but since the sketch is probably static we can see the output in the thumb image so i'll allow it!

PR #1892, opened 2025-08-24
Adding Robot Controller Live streams

shiffman:
I also removed the requirement of credits, does that perhaps have unintended consequences?


shiffman:
This was failing local tests since the robot-controller streams are referring to each other, but weren't actually part of the build yet (at least I think that's why it failed). Wondering what will happen now....

shiffman:
Also wondering if this merits a new track (especially if we get to a part 3 or just leave them in the livestreams track?)

shiffman:
> Ok, I'm going to temporarily just put in the YouTube urls so that the build will work and then later I can replace with the relative urls.

Never mind about this, i had forgotten to add the videos to the livestreams track page.

shiffman:
I know I requested a bunch of reviews, but I'm going to go ahead and merge this to get it on the website, I welcome new PRs or issues as improvements!

fturmel:
@shiffman it should be OK to relax the credits validation rule if that's the intent. Did you want to have the option to not include the credits key at all? Looking at the build failure I think we'd just need to remove the required rule in the graphql schema. LMK if validation rules were wrong or need to be adjusted further, happy to help.

shiffman:
> @shiffman it should be OK to relax the credits validation rule if that's the intent. Did you want to have the option to not include the credits key at all? Looking at the build failure I think we'd just need to remove the required rule in the graphql schema. LMK if validation rules were wrong or need to be adjusted further, happy to help.

I'm fine with the current status of requiring it but allowing an empty array. . but open to better suggestions!

PR #1908, opened 2025-09-15
Showcase Submission for challenges/171-wave-function-collapse

ch00ooch00oo:
Thank you HKH for your contribution! A member of the Coding Train team will review it shortly.

* [Diagonal Lines](http://ozom.dk/WFC/Wave_Function_Collpase_CS.rar)
* HKH

preview image


shiffman:
I checked the `rar` file and it is C# code, I would prefer this to be a link to a github repo but there is nothing malicious in the file so it's ok to keep. The screenshot is lovely! I'm goign to close #1909 and #1910 however, as I don't think it makes sense to have 3 variations as 3 separate submissions.

PR #1989, opened 2026-08-04
Fix import assertions to import attributes syntax

fturmel:
Hey @shiffman, hope all is well.

I noticed an error in the Netlify pipeline when you merged your PR yesterday. This should fix the problem.

Import assertions were superseded by import attributes after we shipped this YouTube redirect feature on the site. Node 22+ doesn't support this assert syntax anymore. I think we're seeing this issue appear now because Node 20 is EOL, and Netlify packages edge functions to run on Deno Deploy where the actual runtime is Deno AFAIK.

CleanShot 2026-08-04 at 14 51 20@2x

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import/with
https://v8.dev/features/import-attributes


fturmel:
Looks fixed in the preview build.

CleanShot 2026-08-04 at 15 20 00@2x

shiffman:
Thanks for catching this!!

PR #1990, opened 2026-08-07
Improve SVG handling

fturmel:
I reopened the source code for the website a few days ago and noticed I had been sitting on some SVG related optimizations that were never finished/merged.

This essentially replaces the `gatsby-plugin-react-svg` with `@svgr/webpack` so SVGs are automatically optimized by `svgo`. It's going to allow upgrading to newer React versions too, there were compat problems with that abandoned Gatsby plugin. The "*.mini.svg" files were removed, as well as some dupes.

The heavy hero train SVG on the homepage is also now 95% smaller (1.6mb to 90kb), which should make the page load faster and use up less bandwidth. It's still an SVG, but the large embedded PNG image inside it was scaled down and converted to a WebP image.

(I'm going to attempt to update the project dependencies over the coming days as well, watch out for another PR soon. Latest Gatsby version should be able to run on Node 24 which is the current LTS).


shiffman:
Thanks for working on this, so great!!

